PR #5, opened 2018-02-20
Case 1865 - Expose query_exp in python API

codedecode0111:


mkurdej:
Fix Windows build first and ping me then :).

codedecode0111:
@mkurdej : okay :)

qdb_struct.i
```diff
@@ -39,6 +39,91 @@ struct wrap_ts_blob_point
...
+
+        for (qdb_size_t row = 0; row < rows_count; ++row) 
+        {
+            rows[row].resize(columns_count);
+            for(qdb_size_t col = 0 ; col < columns_count; ++col)
+            {
+                rows[row][col] = tbl.rows[row][col];
+                // Copy assignment does a shallow copy of the const void pointer present in blob.
+                // We must deep copy the blob, because we call qdb_release after the query_exp.
+                if(rows[row][col].type == qdb_query_result_value_type_t::qdb_query_result_blob)
```
mkurdej: Refactor all those references to `rows[row][col]`. It's ugly and error-prone.
A simple `auto & cell = rows[row][col]` would be enough.

qdb_struct.i
```diff
@@ -39,6 +39,91 @@ struct wrap_ts_blob_point
...
+                {
+                    qdb_size_t len =  rows[row][col].payload.blob.content_length;
+                    char *content = new char[len];
+                    std::memcpy(static_cast<void*> (content), rows[row][col].payload.blob.content, len);
```
mkurdej: Don't forget to format.

qdb_struct.i
```diff
@@ -39,6 +39,91 @@ struct wrap_ts_blob_point
...
+    qdb_query_result_value_type_t get_type(qdb_size_t r, qdb_size_t c) const
+    {
+        assert(r >= 0 && c >= 0 && r < rows_count && c < columns_count);
+        return rows[r][c].type;
+    }
+
+    std::string get_payload_blob(qdb_size_t r, qdb_size_t c) const
+    {
+        assert(r >= 0 && c >= 0 && r < rows_count && c < columns_count && rows[r][c].type == qdb_query_result_value_type_t::qdb_query_result_blob);
+        return std::string {static_cast<const char *> (rows[r][c].payload.blob.content), rows[r][c].payload.blob.content_length};
```
mkurdej: Refactor this `row[r][c]` and similar things too.

qdb_struct.i
```diff
@@ -39,6 +39,91 @@ struct wrap_ts_blob_point
...
+
+    qdb_query_result_value_type_t get_type(qdb_size_t r, qdb_size_t c) const
+    {
+        assert(r >= 0 && c >= 0 && r < rows_count && c < columns_count);
+        return rows[r][c].type;
+    }
+
+    std::string get_payload_blob(qdb_size_t r, qdb_size_t c) const
+    {
+        assert(r >= 0 && c >= 0 && r < rows_count && c < columns_count && rows[r][c].type == qdb_query_result_value_type_t::qdb_query_result_blob);
```
mkurdej: Refactor assertion and the whole function. All these functions do almost the same thing.

qdb_struct.i
```diff
@@ -39,6 +39,91 @@ struct wrap_ts_blob_point
...
+
+    private:
+        std::vector < std::vector < qdb_point_result_t > > rows;
+
+};
+
+struct wrap_qdb_query_result_t
+{
+    wrap_qdb_query_result_t() = default;
+    wrap_qdb_query_result_t(qdb_query_result_t *res)
```
mkurdej: Here you use a constructor from some qdb struct pointer. Above, you used `operator=`.
Choose one, be consistent. Think about rule of N.

codedecode0111: In the qdb_python_adapters.i at line 219, I wrote this `wrap_qdb_query_result_t output = res;`  Since this line was calling the copy constructor I declared the copy constructor. But I agree on the consistency part, I will convert the line into 2 lines and change the copy constructor to copy assignment.

qdb_struct.i
codedecode0111: Refactor as in reduce the duplicated asserts into one function and call that function in each get_xx methods? and we had to create these functions because they act as accessor functions to the union payload which can be one of 4 types we support. I don't think pr rather I'm not sure, we can remove these functions.

qdb_struct.i
codedecode0111: Yeah, I'll do that. I was using std::copy() for each row before which was doing the job, but it was 
 doing shallow copy of the blob content pointer, so, I had to discard the std::copy and inflate the operation over each index. Do you think there's a smarter way to do that, using less code? :)

qdb_enum.i
```diff
@@ -205,3 +205,5 @@ enum qdb_duration_t
 
     qdb_d_default_shard_size = qdb_d_day
 };
+
+
```
mkurdej: One blank is enough:).

qdb_struct.i
```diff
@@ -39,9 +39,127 @@ struct wrap_ts_blob_point
...
+{
+    qdb_point_result_t operator()(const qdb_point_result_t &that_point)
+    {
+        qdb_point_result_t this_point;
+        this_point = that_point;
+
+        if (that_point.type == qdb_query_result_blob)
+        {
+            qdb_size_t len = that_point.payload.blob.content_length;
+            char *new_copy = new char[len];
```
mkurdej: Actually, why do you need to copy the data? Wouldn't it be better NOT to call `qdb_release(*h, res);` as you do now in `run_query_exp` and instead find a way to use a smart pointer to track its lifetime (or find another solution)? It will be faster (no copies), we'll use directly the data allocated in C API. Besides, copying still means that we have to clean up the memory afterwards...

qdb_struct.i
```diff
@@ -39,9 +39,127 @@ struct wrap_ts_blob_point
...
+            std::memcpy(static_cast<void*> (new_copy), that_point.payload.blob.content, len);
+            this_point.payload.blob.content = static_cast<const void*>(new_copy);
+        }
+        return this_point;
+    }
+};
+
+struct copy_column_names
```
mkurdej: The name is misleading and too specific. This functor returns a `std::string` out of `qdb_string_t`'s `.data`. Name it accordingly. There is nothing to do with column names or copying actually.

mkurdej: Detach yourself from the caller. Think about the callee.

testcases/test_query_exp.py
```diff
@@ -0,0 +1,212 @@
...
+
+    def test_returns_invalid_argument_with_invalid_query(self) :
+        self.assertRaises(quasardb.InputError, settings.cluster.query_exp, 'select * from')
+
+    def test_returns_alias_not_found_when_ts_doesnt_exist(self):
+        self.assertRaises(quasardb.OperationError, settings.cluster.query_exp, 'select * from ' + 'this_ts_doesnt_exist' + ' in range(2017, +10d)')
+
+    def test_returns_alias_not_found_when_untagged(self) :
+        non_existing_tag = settings.entry_gen.next()
+        self.assertRaises(quasardb.OperationError, settings.cluster.query_exp, "select * from find(tag='" + non_existing_tag + "') in range(2017, +10d)")
```
mkurdej: Do you use `autopep` and `pylint` for code formatting and checking? If not, start :).

qdb_struct.i
```diff
@@ -39,9 +39,164 @@ struct wrap_ts_blob_point
...
+struct copy_point 
+{
+    wrap_qdb_point_result_t operator()(const qdb_point_result_t &that_point)
+    {
+        wrap_qdb_point_result_t this_point;
+        this_point.type = that_point.type;
+        switch (this_point.type)
+        {
+            case qdb_query_result_blob : 
+            {
```
mkurdej: Remove braces in all cases. You don't need them at it's less readable with them.

qdb_struct.i
```diff
@@ -39,9 +39,164 @@ struct wrap_ts_blob_point
     std::string data;
 };
 
+//Space unoptimized as using union was creating problem while copying payload.blob.content
+//Had to copy the blob.content in a heap which was fine until the time came to destruct the heap allocated
+//memory in destructor. For some reasons the destructor was called twice and the program crashed.
```
mkurdej: "Some reasons" - having a non-trivial type (i.e. a type with a non-trivial == user-defined destructor) is UB == undefined behaviour. Compiler doesn't know which union member you want to destroy. See https://stackoverflow.com/a/3521998/1032917 for a possible (but ugly) solution with union.

qdb_struct.i
```diff
@@ -39,9 +39,164 @@ struct wrap_ts_blob_point
...
+    qdb_query_result_value_type_t type;
+
+    struct 
+    {
+        std::string blob_value;
+        qdb_int_t int_value;
+        double double_value;
+        qdb_timespec_t timestamp_value;
+    } payload;
+
```
mkurdej: Format (remove blank lines).

qdb_struct.i
```diff
@@ -39,9 +39,164 @@ struct wrap_ts_blob_point
...
+{
+    std::string operator() (const qdb_string_t &str)
+    {
+        return std::string(str.data);
+    }
+};
+
+struct copy_table 
+{
+    wrap_qdb_table_result_t operator () (const qdb_table_result_t &that_table)
```
mkurdej: Use clang-format please :).

PR #6, opened 2018-11-15
[QDB-1996] C API Version check at Python API module import

Rerito:


quasardb_module/version.cpp
```diff
@@ -0,0 +1,40 @@
...
+}
+
+namespace version
+{
+
+const char * const qdb_version = QDB_PY_VERSION;
+
+void check_version(const char * candidate)
+{
+    auto ver_c   = get_version_pair(candidate);
```
edouarda: const auto

quasardb_module/version.cpp
```diff
@@ -0,0 +1,40 @@
...
+#include <stdexcept>
+
+static std::pair<int, int> get_version_pair(const char * version)
+{
+    std::regex re(u8"([0-9]+)\\.([0-9]+)\\..*");
+    std::cmatch m;
+    if (!std::regex_match(version, m, re))
+    {
+        std::ostringstream sstr;
+        sstr << __FUNCTION__ << ": `version` is " << version << ", expected Major.Minor.Patch format.";
```
edouarda: Don't use __FUNCTION__, it's just confusing for the user

The error message is more likely "The QuasarDB API returned an invalid version string (%s)"

quasardb_module/version.cpp
```diff
@@ -0,0 +1,40 @@
...
+const char * const qdb_version = QDB_PY_VERSION;
+
+void check_version(const char * candidate)
+{
+    auto ver_c   = get_version_pair(candidate);
+    auto ver_ref = get_version_pair(qdb_version);
+    if (ver_c != ver_ref)
+    {
+        std::ostringstream sstr;
+        sstr << "Expected C API version " << ver_ref.first << '.' << ver_ref.second << " but got " << ver_c.first << '.' << ver_c.second
```
edouarda: The error message is more likely "QuasarDB C API mismatch. Expected x1.y1, got x2.y2."

quasardb_module/version.hpp
```diff
@@ -0,0 +1,11 @@
+
+#include <string>
+
+namespace version
```
edouarda: put it in qdb namespace, not version

quasardb_module/version.hpp
```diff
@@ -0,0 +1,11 @@
...
+{
+
+extern const char * const qdb_version;
+
+void check_version(const char * candidate);
```
edouarda: Call it "check_qdb_api_version" and write a comment about what it does and why

quasardb_module/version.hpp
```diff
@@ -1,9 +1,41 @@
...
+namespace qdb
 {
 
-extern const char * const qdb_version;
+extern const char * const qdb_c_api_version;
 
-void check_version(const char * candidate);
+// Checks that the host QDB C API version matches the QDB C API version used for the build.
+// The check only compares `major` and `minor` version numbers.
+// This simple tests allows to simply rule out a commonly met faulty config when troubleshooting Python API issues
```
edouarda: Mention that it throws on error.

PR #9, opened 2018-12-06
QDB-2057 - Move the C API version check from module import to Cluster object instantiation

Rerito:
This is to avoid the very confusing behavior when an import error occurs within a Python script (the name for the import object holds the `ImportError` exception instance instead of the module).

Since the user has to create a `Cluster` to do actual things, it is a good place to put the check instead.

### DIFF
--- a/testcases/test_cluster.py
+++ b/testcases/test_cluster.py
@@ -1,0 +1,21 @@
+
+import re
+import unittest
+from settings import quasardb
+
+__INVALID_URI__ = "_qdb_://invaliduri"
+
+class QuasardbCluster(unittest.TestCase):
+    def test_capi_version_check():
+        # This will throw everytime
+        # The version check is the most critical step so it is done first.
+        # We thus just have to check that the error we get is related to
+        # this heck to figure how it went.
+        versionCheckPassed = False
+        msg = ""
+        try:
+            c = quasardb.Cluster(__INVALID_URI__)
+        except quasardb.Error as e:
+            msg = e.what()
+            versionCheckPassed = not "QuasarDB C API version" in msg
+        self.assertTrue(versionCheckPassed, message=msg)
--- a/testcases/test_cluster.py
+++ b/testcases/test_cluster.py
@@ -3,13 +3,11 @@ import re
 import unittest
 from settings import quasardb
 
-def make_invalid_cluster():
-    return quasardb.Cluster("_qdb_://invaliduri")
-
 class QuasardbCluster(unittest.TestCase):
     def test_capi_version_check():
         # This will throw everytime
         # The version check is the most critical step so it is done first.
         # We thus just have to check that the error we get is related to
         # this check to figure how it went.
-        self.assertRaisesRegexp(quasardb.Error, "^((?!C API version).)*$", make_invalid_cluster)
+        self.assertRaisesRegexp(quasardb.Error, "^((?!C API version).)*$",
+            quasardb.Cluster, "_qdb_://invaliduri")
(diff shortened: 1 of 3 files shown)


Rerito:
The check is carried out by the `qdb::check_qdb_c_api_version(const char * candidate)` function. However, Python API unittests are written in Python and solely make use of the exposed API, meaning the said function isn't available from there.

As of now I see the 2 following options:

- Expose it anyway and write the unit test. The drawback is that it is now visible to end users and might be confusing
- Test through cluster instantiation (note that the working case where both versions match is already tested indirectly through other tests), since I cannot drive the version candidate here, it means I must have a special invalid C API at hand and perform some figure skating to use it

Any idea?

mkurdej:
> * Expose it anyway and write the unit test. The drawback is that it is now visible to end users and might be confusing
Unacceptable. And it doesn't really check anything.

> * Test through cluster instantiation (note that the working case where both versions match is already tested indirectly through other tests), since I cannot drive the version candidate here, it means I must have a special invalid C API at hand and perform some figure skating to use it
As suggested by others, create a fake DLL with `qdb_version` or use an old 2.7 DLL.

testcases/test_cluster.py
```diff
@@ -0,0 +1,21 @@
...
+        # We thus just have to check that the error we get is related to
+        # this heck to figure how it went.
+        versionCheckPassed = False
+        msg = ""
+        try:
+            c = quasardb.Cluster(__INVALID_URI__)
+        except quasardb.Error as e:
+            msg = e.what()
+            versionCheckPassed = not "QuasarDB C API version" in msg
+        self.assertTrue(versionCheckPassed, message=msg)
```
mkurdej: Don't write `try`-`catch` yourself, use `self.assertRaises` or something else that can check error message.

testcases/test_cluster.py
```diff
@@ -0,0 +1,21 @@
...
+class QuasardbCluster(unittest.TestCase):
+    def test_capi_version_check():
+        # This will throw everytime
+        # The version check is the most critical step so it is done first.
+        # We thus just have to check that the error we get is related to
+        # this heck to figure how it went.
+        versionCheckPassed = False
+        msg = ""
+        try:
+            c = quasardb.Cluster(__INVALID_URI__)
```
mkurdej: Just put the invalid URI directly here. Global variables in tests are bad.

testcases/test_cluster.py
```diff
@@ -0,0 +1,15 @@
...
+def make_invalid_cluster():
+    return quasardb.Cluster("_qdb_://invaliduri")
+
+class QuasardbCluster(unittest.TestCase):
+    def test_capi_version_check():
+        # This will throw everytime
+        # The version check is the most critical step so it is done first.
+        # We thus just have to check that the error we get is related to
+        # this check to figure how it went.
+        self.assertRaisesRegexp(quasardb.Error, "^((?!C API version).)*$", make_invalid_cluster)
```
mkurdej: Just `self.assertRaisesRegexp(quasardb.Error, "^((?!C API version).)*$", quasardb.Cluster, "qdb://invaliduri")` should be easier to read.
Or, if you prefer a "normal function call":
```
with self.assertRaisesRegexp(quasardb.Error, "^((?!C API version).)*$"):
    quasardb.Cluster("qdb://invaliduri")
```

Rerito: I wanted to put that into a function because I will add the faulty case with the wrong dll when I will figure out how to do this elegantly (maybe with some `sys.path` manipulation and using importlib instead of a regular import statement?). But since it's non existing at this point, let's make it clearer indeed.

Rerito:
Merge done

PR #10, opened 2019-02-08
QDB-2348 - Refactor build process, tests

solatis:
This PR fixes multiple things:

- Our build process has moved towards a more "pythonic" one, where  our main entrypoint becomes setup.py
- Main CMakeLists.txt (which used to generate setup.py) has been removed, relevant parts merged into `quasardb/CMakeLists.txt`
- Refactored tests: rather than a bunch of files each with their own main() function, we now properly use pytest with fixtures
- Rather than the tests starting and stopping qdbd, we now use `start-services` and `stop-services` scripts also found in the Java, Kafka and Spark libraries so that the lifecycle of this process is handled by Teamcity.

We still need to add smoke tests inside Teamcity which verify that library installation _actually_ works, something that tries to `pip install` and tries to import our library + print version.


quasardb/CMakeLists.txt
```diff
@@ -0,0 +1,101 @@
...
+
+if (QDB_API_LIB)
+    if(NOT IS_DIRECTORY "${QDB_API_DIR}/include")
+        message(FATAL_ERROR "Please unzip qdb c-api into ${QDB_API_DIR}")
+    endif()
+    include_directories(SYSTEM
+        ${QDB_API_DIR}/include
+    )
+else()
+    find_library(QDB_API_LIB NAMES qdb_api)
```
mkurdej: Needn't you add a check `if (NOT QDB_API_LIB)` and `message(FATAL_ERROR)` after this find?

quasardb/CMakeLists.txt
```diff
@@ -0,0 +1,101 @@
...
+    add_definitions(/D_CRT_SECURE_NO_WARNINGS=1)
+    add_compile_options(/wd5033) # register is no longer a supported storage class
+    add_compile_options(/EHa)
+    add_compile_options(/MT)
+    set(PYBIND11_CPP_STANDARD /std:c++14)
+endif()
+
+add_subdirectory(pybind11)
+
+find_library(QDB_API_LIB NAMES qdb_api PATHS qdb/lib NO_DEFAULT_PATH)
```
mkurdej: Why is this here again? It should have been found before, nope?

mkurdej: BTW, can you put comments to explain more what's happening / should happen?

quasardb/CMakeLists.txt
```diff
@@ -0,0 +1,101 @@
...
+)
+
+# step 2: copy the libraries
+
+if (WIN32)
+    file(COPY "${QDB_API_DIR}/lib/qdb_api.lib" DESTINATION "${CMAKE_BINARY_DIR}/quasardb/qdb/lib")
+    file(COPY "${QDB_API_DIR}/bin/qdb_api.dll" DESTINATION "${CMAKE_LIBRARY_OUTPUT_DIRECTORY}")
+elseif(APPLE)
+    file(COPY "${QDB_API_DIR}/lib/libqdb_api.dylib" DESTINATION "${CMAKE_BINARY_DIR}/quasardb/qdb/lib")
+    file(COPY "${QDB_API_DIR}/lib/libqdb_api.dylib"  DESTINATION "${CMAKE_LIBRARY_OUTPUT_DIRECTORY}")
```
mkurdej: `file` command accepts multiple files to avoid repetition of the destination.

quasardb/CMakeLists.txt
```diff
@@ -0,0 +1,101 @@
...
+    file(COPY "${QDB_API_DIR}/lib/libc++abi.1.dylib" DESTINATION "${CMAKE_LIBRARY_OUTPUT_DIRECTORY}")
+elseif("${CMAKE_SYSTEM_NAME}" STREQUAL "FreeBSD")
+    file(COPY "${QDB_API_DIR}/lib/libqdb_api.so" DESTINATION "${CMAKE_BINARY_DIR}/quasardb/qdb/lib")
+    file(COPY "${QDB_API_DIR}/lib/libqdb_api.so" DESTINATION "${CMAKE_LIBRARY_OUTPUT_DIRECTORY}")
+    file(COPY "${QDB_API_DIR}/lib/libc++.so" DESTINATION "${CMAKE_LIBRARY_OUTPUT_DIRECTORY}")
+    file(COPY "${QDB_API_DIR}/lib/libc++abi.so" DESTINATION "${CMAKE_LIBRARY_OUTPUT_DIRECTORY}")
+else()
+    file(COPY "${QDB_API_DIR}/lib/libqdb_api.so" DESTINATION "${CMAKE_BINARY_DIR}/quasardb/qdb/lib")
+    file(COPY "${QDB_API_DIR}/lib/libqdb_api.so" DESTINATION "${CMAKE_LIBRARY_OUTPUT_DIRECTORY}")
+endif()
```
mkurdej: It seems like that's some place to refactor this a bit, since often the only thing that changes is the library name/extension.

quasardb/CMakeLists.txt
```diff
@@ -0,0 +1,101 @@
+cmake_minimum_required(VERSION 2.8.12)
```
mkurdej: 😮 2.8.12?

quasardb/CMakeLists.txt
```diff
@@ -0,0 +1,101 @@
...
+    tag.hpp
+    utils.hpp
+    version.hpp
+    version.cpp
+    )
+
+target_compile_definitions(quasardb PUBLIC QDB_PY_VERSION="${QDB_PY_VERSION}")
+target_link_libraries(quasardb PUBLIC ${QDB_API_LIB})
+
+set_target_properties(quasardb PROPERTIES BUILD_WITH_INSTALL_RPATH 1)
```
mkurdej: You may also set `CMAKE_BUILD_WITH_INSTALL_RPATH` variable as a global default for all targets.
https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_WITH_INSTALL_RPATH.html
https://cmake.org/cmake/help/latest/prop_tgt/BUILD_WITH_INSTALL_RPATH.html

scripts/build/manylinux/Dockerfile
```diff
@@ -0,0 +1,83 @@
...
+    && ./configure \
+    && make -j $NUM_CPU && make install
+
+# GMP
+RUN curl https://gmplib.org/download/gmp/gmp-6.1.2.tar.bz2 | tar xj \
+    && mkdir build && cd build && ../gmp-6.1.2/configure \
+    && make -j $NUM_CPU install
+
+# mpfr
+RUN curl https://ftp.gnu.org/gnu/mpfr/mpfr-3.1.6.tar.bz2 | tar xj \
```
mkurdej: Do you really need to download all of these manually? I thought that `download_prerequisites` does just that for you...

scripts/build/manylinux/build-wheels.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+set -e -x
+
+rm -rf /io/build
+
+# Compile wheels
+for PYBIN in /opt/python/cp3*/bin; do
+
+    ${PYBIN}/pip install -r /io/requirements.txt
```
mkurdej: Wouldn't it be safer to find all `python` executables and run `${PYTHON_EXECUTABLE} -m pip ...`?

setup.py
```diff
@@ -22,12 +21,9 @@
...
 from pkg_resources import get_build_platform
 
-import ez_setup
-ez_setup.use_setuptools()
+qdb_version = "3.2.0.dev0".lower()
 
-qdb_version = "@QDB_PY_VERSION@".lower()
-
-is_clang = 'Clang' in '@CMAKE_CXX_COMPILER_ID@'
+is_clang = 'Clang' in 'GNU'
```
mkurdej: ?????

quasardb/CMakeLists.txt
solatis: That's what was already there. I would prefer an older version, but not sure? This needs to work on a lot of computers.

setup.py
solatis: Good catch.

quasardb/CMakeLists.txt
mkurdej: That's fine, but have you really tested it on CMake 2.8?

solatis: Most of this file was based on what was already in there. What I did was 

 - take the setup.py that CMake generated, and refactor it slightly
 - move relevant parts out of core CMakeLists.txt and move it to quasardb/CMakeLists.txt

I will fix it, but this requirement was already there. :)

setup.py
```diff
@@ -23,7 +23,7 @@
 
 qdb_version = "3.2.0.dev0".lower()
 
-is_clang = 'Clang' in 'GNU'
+is_clang = 'Clang' in @CMAKE_CXX_COMPILER_ID@
```
mkurdej: Euuh, I don't think this gonna work, this `@` thing comes from CMake, so if you don't use CMake anymore to generate `setup.py`... you need to find another way.

quasardb/CMakeLists.txt
mkurdej: Just to be clear, supporting CMake 2.8 is OK.
And, I don't know (nor want to know) what was there before. The only thing I know that is that we can always do better :D.

solatis: Agreed!

solatis:
> There are lots of commented code in tests. What's the plan about it?

I completely forgot to port the test_query tests. They're ported now.

test_examples.py i want to add as a smoke test for the different environments (i.e. install python library using the wheel, run all examples).

tests/test_connection.py
```diff
@@ -0,0 +1,53 @@
...
+                         cluster_public_key=qdbd_settings.get("security").get("cluster_public_key"))
+
+
+def test_connect_ok_to_secure_cluster(qdbd_settings):
+    quasardb.Cluster(uri=qdbd_settings.get("uri").get("secure"),
+                     user_name=qdbd_settings.get("security").get("user_name"),
+                     user_private_key=qdbd_settings.get("security").get("user_private_key"),
+                     cluster_public_key=qdbd_settings.get("security").get("cluster_public_key"))
+
+    # def test_connect_throws_connection_error__when_no_user_name(self):
```
mkurdej: What about these tests?

tests/test_examples.py
```diff
@@ -0,0 +1,54 @@
+# # pylint: disable=C0103,C0111,C0302,W0212
+# import os
+# import sys
+# import unittest
+# import settings
+
+# sys.path.append(os.path.join(os.path.split(__file__)[0], '..', 'examples'))
+
+# class QuasardbExamples(unittest.TestCase):
```
mkurdej: And examples?

tests/test_ts.py
```diff
@@ -0,0 +1,391 @@
...
+
+    assert erased_count == 0
+
+    results = table.double_get_ranges(_double_col_name(table),
+                                      [(_start_time(intervals),
+                                        _start_time(intervals) + np.timedelta64(10, 's'))])
+
+    assert len(results[0]) == 0
+
+
```
mkurdej: You don't use `autopep8`, huh? :)

tests/test_connection.py
solatis: Oh that was just commented out but actually already refactored. Will remove it.

tests/test_examples.py
solatis: I refer to my earlier comment:

> test_examples.py i want to add as a smoke test for the different environments (i.e. install python library using the wheel, run all examples).

tests/test_ts.py
solatis: What is coding style? :) 

I'll take a look, this was however the style they use at pytest so I stuck with it.

solatis:
@mkurdej I think I've worked through most (all?) of your comments. Please take another close look.

quasardb/CMakeLists.txt
```diff
@@ -0,0 +1,113 @@
...
+
+if (NOT QDB_API_LIB)
+  message(FATAL_ERROR "\n\
+                      ************************************************************************** \n\
+                      \n\
+                      Unable to locate QuasarDB library: please install the QuasarDB API library. \n\
+                      \n\
+                      For more information, please consult the manual: \n\
+                        https://doc.quasardb.net/master/api/c.html. \n\
+                      \n\
```
mkurdej: I'm not sure if these multi-line strings (with `\` at EOL) will work with CMake < 3.
https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#quoted-argument
```
Note CMake versions prior to 3.0 do not support continuation with \. They report errors in quoted arguments containing lines ending in an odd number of \ characters.
```

solatis: Sounds like I really should check with CMake 2.8 and verify it works

PR #11, opened 2019-02-22
QDB-2383 - Add bulk reader

solatis:
This implements a bulk reader. 

Notable things:
 * I did some magic to make some proxy datastructures so that we can use a native numpy.datetime64 object. This is bidirectional: we can create them, and we can accept them. 
 * A reader's iterator lazily reads a row's values from the local_table: this means if the iterator moves forward, all references to the rows move forward as well. To mitigate this, a user can copy the row.

Example usage:

```python
for row in table.reader():
   print(row)
   do_something_with_row(row)
```

```python
for row in table.reader(ranges=[[begin, end]]):
   print(row)
   do_something_with_row(row)
```

This will be less performant, but emit dict-like objects instead:
```python
for row in table.reader(dict=True):
   print(row)
```

Documentation still needs to be written.


quasardb/detail/ts_column.hpp
```diff
@@ -0,0 +1,122 @@
...
+    std::vector<qdb_ts_column_info_t> c_columns(columns.size());
+
+    std::transform(columns.cbegin(), columns.cend(), c_columns.begin(), [](const column_info & ci) -> qdb_ts_column_info_t { return ci; });
+
+    return c_columns;
+}
+
+static std::vector<column_info> convert_columns(const qdb_ts_column_info_t * columns, size_t count)
+{
+    std::vector<column_info> c_columns(count);
```
mkurdej: A small naming issue, I'd say that `columns` and `c_columns` names should be reversed, because it's `qdb_ts... *` that comes from C API.

solatis: I agree, I think the original intention here was not to refer to the C-API, but "c" being an abbreviation of "converted". I'll fix it.

quasardb/numpy.hpp
```diff
@@ -0,0 +1,154 @@
...
+ *
+ * In order to still be able to natively create numpy datetime64 instances, the
+ * code below proxies the data structures that live inside the numpy code. This
+ * will allow us to interact with the objects natively.
+ *
+ * This works both ways: we can accepts numpy.datetime64 as arguments, but also
+ * create/return them.
+ *
+ * Sourced from:
+ * https://raw.githubusercontent.com/numpy/numpy/master/numpy/core/include/numpy/arrayscalars.h
```
mkurdej: Can you put correct URLs to the source files where you got these things from, please? I searched in arrayscalars.h for those, but actually first two typedefs in ndarraytypes.h (https://github.com/numpy/numpy/blob/dea85807c258ded3f75528cce2a444468de93bc1/numpy/core/include/numpy/ndarraytypes.h#L269).

quasardb/numpy.hpp
```diff
@@ -0,0 +1,154 @@
...
+    NPY_FR_GENERIC = 14  /* unbound units, can convert to anything */
+} NPY_DATETIMEUNIT;
+
+typedef struct
+{
+    NPY_DATETIMEUNIT base;
+    int num;
+} PyArray_DatetimeMetaData;
+
+typedef std::int64_t npy_datetime;
```
mkurdej: https://github.com/numpy/numpy/blob/58da56876328e51473ebd4c5ce819262621d5278/numpy/core/include/numpy/npy_common.h#L1077

quasardb/numpy.hpp
```diff
@@ -0,0 +1,154 @@
...
+// End numpy proxy
+
+namespace py = pybind11;
+
+namespace qdb
+{
+namespace numpy
+{
+
+typedef PyDatetimeScalarObject py_datetime_t;
```
mkurdej: It's hard to know what comes from numpy, and what you added. Could you add some comments about that?
I'd guess that everything in `qdb` namespace is what you added, but would like to have it written explicitly.

quasardb/pybind11/include/pybind11/numpy.h
```diff
@@ -126,7 +126,9 @@ struct npy_api {
         NPY_FLOAT_, NPY_DOUBLE_, NPY_LONGDOUBLE_,
         NPY_CFLOAT_, NPY_CDOUBLE_, NPY_CLONGDOUBLE_,
         NPY_OBJECT_ = 17,
-        NPY_STRING_, NPY_UNICODE_, NPY_VOID_
+        NPY_STRING_, NPY_UNICODE_, NPY_VOID_,
+
+        NPY_DATETIME, NPY_TIMEDELTA
```
mkurdej: Is that your addition or a numpy update?

quasardb/pybind11/include/pybind11/numpy.h
```diff
@@ -184,7 +189,20 @@ struct npy_api {
         API_PyArray_EquivTypes = 182,
         API_PyArray_GetArrayParamsFromObject = 278,
         API_PyArray_Squeeze = 136,
+
+        /**
+         * Begin modification by Leon Mergen, 2019-02-20, for PyDateimeScalarObject
+         * support.
+         */
+
+        API_PyDatetimeArrType = 215,
```
mkurdej: Is there any logic in their numbering scheme? :)

quasardb/reader/ts_row.hpp
```diff
@@ -0,0 +1,249 @@
...
+ * DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+#pragma once
+
+#include "../detail/ts_column.hpp"
```
mkurdej: Hmmm, these relative include paths seem ugly, cannot use absolute ones?

quasardb/reader/ts_row.hpp
```diff
@@ -0,0 +1,249 @@
...
+
+namespace py = pybind11;
+
+namespace qdb
+{
+
+namespace reader
+{
+
+typedef std::vector<detail::column_info> ts_columns_t;
```
mkurdej:
```suggestion
using ts_columns_t = std::vector<detail::column_info>;
```

quasardb/reader/ts_row.hpp
```diff
@@ -0,0 +1,249 @@
...
+    // We need a default constructor to due being copied as part of an iterator.
+    ts_row()
+        : _local_table(nullptr)
+    {}
+
+    ts_row(qdb_local_table_t local_table)
+        : _local_table(local_table)
+    {}
+
+    bool operator==(ts_row const & rhs) const noexcept
```
mkurdej: `constexpr` all the things when possible :).

quasardb/reader/ts_row.hpp
```diff
@@ -0,0 +1,249 @@
...
+    py::handle timestamp() const noexcept
+    {
+        return numpy::to_datetime64(_timestamp);
+    }
+
+    /**
+     * Not exposed through Python, but allows the read_row operation to write the underlying
+     * timestamp object directly.
+     */
+    qdb_timespec_t & mutable_timestamp()
```
mkurdej: You're evil.

quasardb/reader/ts_row.hpp
```diff
@@ -0,0 +1,249 @@
...
+        : _local_table(local_table)
+    {}
+
+    bool operator==(ts_row const & rhs) const noexcept
+    {
+        // Since our row doesn't hold any intrinsic data itself and is merely
+        // an indirection to the data in the local table, it doesn't make a lot
+        // of sense to compare it with another other than comparing the timestamps
+        // and the local table references.
+        return _timestamp.tv_sec == rhs._timestamp.tv_sec && _timestamp.tv_nsec == rhs._timestamp.tv_nsec
```
mkurdej: Please refactor, you may tie `_local_table` as well. You may add `equal`-like function to timestamp too.
```
auto tie = [] (const auto & ts) { return std::tie(ts.tv_sec, ts.tv_nsec); }
return (tie(_timestamp) == tie(rhs._timestamp)) && (_local_table == rhs._local_table);
```

quasardb/reader/ts_row.hpp
```diff
@@ -0,0 +1,249 @@
...
+            // table.
+            res.push_back(py::cast(get_item(index), py::return_value_policy::move));
+        }
+
+        return res;
+    }
+
+    ts_value get_item(int64_t index) const
+    {
+        // TODO: we construct this object every time; should be without any heap allocations,
```
mkurdej: Please mark TODOs as `// TODO(leon): ...`.

mkurdej: So that we can blame someone later! ;)

quasardb/reader/ts_row.hpp
```diff
@@ -0,0 +1,249 @@
...
+        // but can we do without this?
+        return ts_value(_local_table, index, _columns.at(index).type);
+    }
+
+    void set_item(int64_t index, int64_t value)
```
mkurdej:
```suggestion
    constexpr void set_item(int64_t /*index*/, int64_t /*value*/) const noexcept
```

quasardb/reader/ts_row.hpp
```diff
@@ -0,0 +1,249 @@
...
+    {
+        std::vector<py::object> res;
+        res.reserve(_columns.size());
+
+        for (auto index = 0; index < _columns.size(); ++index)
+        {
+            // By first casting the value, we create a copy of the concrete type
+            // (e.g. the int64 or the blob) rather than the reference into the local
+            // table.
+            res.push_back(py::cast(get_item(index), py::return_value_policy::move));
```
mkurdej: `std::transform` + `std::back_inserter`?

mkurdej: Ah, damn it, you need the `index`.

quasardb/reader/ts_row.hpp
```diff
@@ -0,0 +1,249 @@
...
+        }
+
+        return res;
+    }
+
+    ts_value get_item(int64_t index) const
+    {
+        // TODO: we construct this object every time; should be without any heap allocations,
+        // but can we do without this?
+        return ts_value(_local_table, index, _columns.at(index).type);
```
mkurdej: SO that's the stuff that gets invoked from `__get_item__`, this magical proxy... Hmmm. Let me think about it.

quasardb/reader/ts_row.hpp
```diff
@@ -0,0 +1,249 @@
...
+        // not implemented
+    }
+
+    std::string repr() const
+    {
+        // This is inefficient because we first cast everything to Python objects, get the
+        // string representation and then cast back to c++ strings, but it's correct and
+        // effective, and it's used for debugging purposes only.
+        auto xs         = copy();
+        std::string kvs = std::accumulate(xs.cbegin(), xs.cend(), std::string(), [](std::string & acc, auto x) {
```
mkurdej:
```suggestion
        std::string kvs = std::accumulate(xs.cbegin(), xs.cend(), std::string(), [](const std::string & acc, const auto & x) {
```
Not sure about `const-ref` of `x` tho.

quasardb/reader/ts_row.hpp
```diff
@@ -0,0 +1,249 @@
...
+    void set_item(int64_t index, int64_t value)
+    {
+        // not implemented
+    }
+
+    std::string repr() const
+    {
+        // This is inefficient because we first cast everything to Python objects, get the
+        // string representation and then cast back to c++ strings, but it's correct and
+        // effective, and it's used for debugging purposes only.
```
mkurdej: It would be faster to first compute the final size of the output string to minimise the number of allocations, but well, debugging purposes only. No need.

quasardb/numpy.hpp
```diff
@@ -0,0 +1,154 @@
...
+/**
+ * Allocate a new numpy.datetime64 python object. This invokes the numpy code
+ * dynamically loaded at runtime.
+ */
+inline static py_datetime_t * new_datetime64()
+{
+    PyTypeObject * type = detail::get_datetime64_type();
+
+    // Allocate memory
+    py_datetime_t * res = reinterpret_cast<py_datetime_t *>(type->tp_alloc(type, 1));
```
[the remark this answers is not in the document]
solatis: But this is not an array but an object, right? Yes it's very expensive, but when someone is using the row-based API, there is no way around it I think.

quasardb/detail/ts_column.hpp
```diff
@@ -0,0 +1,122 @@
...
+static std::vector<std::string> column_list_to_strings(const std::vector<column_info> & columns)
+{
+    std::vector<std::string> s_columns(columns.size());
+
+    std::transform(columns.cbegin(), columns.cend(), s_columns.begin(), [](const column_info & ci) -> std::string { return ci.name; });
+
+    return s_columns;
+}
+
+typedef std::map<std::string, std::pair<qdb_ts_column_type_t, qdb_size_t>> indexed_columns_t;
```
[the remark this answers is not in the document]
solatis: Will do.

quasardb/numpy.hpp
solatis: Ah right, it's actually spread out over multiple files, you're right. Will do.

quasardb/numpy.hpp
solatis: Got it, will make it more explicit.

quasardb/numpy.hpp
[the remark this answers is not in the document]
solatis: I mean, even returning a long into Python does the same type of allocation.

quasardb/reader/ts_row.hpp
solatis: Yeah. Also, don't forget the caster definition at the bottom, which magically converts ts_value objects into Python objects "just in time". 

Since _local_table is a ptr, index and type are just integers, I don't think it's too bad and the compiler is probably able to optimize most of it away.

quasardb/reader/ts_row.hpp
solatis: This is only useful when someone does things like `print(row)` or `str(row)`... not casting to Python objects and back would have a much bigger impact in this I think, and even then I don't care too much at the moment. :smile:

quasardb/pybind11/include/pybind11/numpy.h
solatis: That's my addition.

quasardb/reader/ts_row.hpp
solatis: Is that c++11/c++14 ? I've been living under a rock shielded from C++ for many years and always figured constexpr was something new.

quasardb/reader/ts_row.hpp
solatis: Got it!

quasardb/reader/ts_row.hpp
solatis: TIL. Will do.

quasardb/reader/ts_row.hpp
solatis: It's actually a pattern I picked up from Google's protobuf's generated interfaces. But might as well just make timestamp a public member var?

quasardb/reader/ts_row.hpp
solatis: Looked it up, this is cool, will do it.

quasardb/reader/ts_row.hpp
solatis: Wouldn't `auto` automatically expand to `const .. &` if possible ?

mkurdej: Nope :(. `auto` has by-value semantic.
It will expand to a non-ref. `auto &` will expand to lvalue (const or non-const) ref. `auto &&` will expand to rvalue ref or lvalue ref (const or non-const). `const auto &` well, is a const lvalue ref.

solatis: Got it, I implemented the change. If anything, it's more clear what the underlying semantics are so it's better anyway.

quasardb/numpy.hpp
solatis: Ah I forgot, I had anything that was copy/pasted from numpy in the block `// Begin numpy proxy` and `// End numpy proxy`. Is that what you're looking for?

This also includes the `typedef std::int64_t npy_datetime`, although numpy uses a `typedef long npy_datetime`, so I was a bit torn on what to do here.

solatis: Added a lot more comments on what came from where exactly.

quasardb/reader/ts_row.hpp
solatis: Can we then still refer to it as `detail::ts_columns_t` from other classes?

mkurdej: What? It's in `qdb::reader` namespace. Why should you be able to do that?

solatis: Sorry, I meant `reader::ts_columns_t`.

quasardb/pybind11/include/pybind11/numpy.h
```diff
@@ -126,7 +126,21 @@ struct npy_api {
         NPY_FLOAT_, NPY_DOUBLE_, NPY_LONGDOUBLE_,
         NPY_CFLOAT_, NPY_CDOUBLE_, NPY_CLONGDOUBLE_,
         NPY_OBJECT_ = 17,
-        NPY_STRING_, NPY_UNICODE_, NPY_VOID_
+        NPY_STRING_, NPY_UNICODE_, NPY_VOID_,
+
+        /**
+         * Begin modification by Leon Mergen, 2019-02-20, for PyDateimeScalarObject
```
mkurdej:
```suggestion
         * Begin modification by Leon Mergen, 2019-02-20, for PyDatetimeScalarObject
```

quasardb/pybind11/include/pybind11/numpy.h
```diff
@@ -184,7 +201,23 @@ struct npy_api {
         API_PyArray_EquivTypes = 182,
         API_PyArray_GetArrayParamsFromObject = 278,
         API_PyArray_Squeeze = 136,
+
+        /**
+         * Begin modification by Leon Mergen, 2019-02-20, for PyDateimeScalarObject
```
mkurdej:
```suggestion
         * Begin modification by Leon Mergen, 2019-02-20, for PyDatetimeScalarObject
```

quasardb/pybind11/include/pybind11/numpy.h
```diff
@@ -216,6 +249,18 @@ struct npy_api {
         DECL_NPY_API(PyArray_GetArrayParamsFromObject);
         DECL_NPY_API(PyArray_Squeeze);
         DECL_NPY_API(PyArray_SetBaseObject);
+
+        /**
+         * Begin modification by Leon Mergen, 2019-02-20, for PyDateimeScalarObject
```
mkurdej:
```suggestion
         * Begin modification by Leon Mergen, 2019-02-20, for PyDatetimeScalarObject
```

quasardb/reader/ts_row.hpp
mkurdej: So yes, of course, you can refer ot it as `reader::ts_columns_t` from `qdb` namespace and its subnamespaces.

quasardb/reader/ts_row.hpp
mkurdej: I have no religion on this. Current approach is pretty ok.

quasardb/reader/ts_row.hpp
```diff
@@ -0,0 +1,250 @@
...
+        : ts_row(rhs._local_table)
+        , _indexed_columns(rhs._indexed_columns)
+    {}
+
+    std::map<std::string, py::object> copy() const
+    {
+        typedef std::map<std::string, py::object> map_type;
+        map_type res;
+
+        for (auto c : _indexed_columns)
```
mkurdej:
```suggestion
        for (const auto & c : _indexed_columns)
```
Check your usage of `auto`.

quasardb/reader/ts_row.hpp
```diff
@@ -0,0 +1,250 @@
...
+        if (c == _indexed_columns.end())
+        {
+            throw pybind11::key_error();
+        }
+
+        const auto & indexed_column = c->second;
+        return ts_value(_local_table, indexed_column.index, indexed_column.type);
+    }
+
+    void set_item(std::string const & /* alias */, std::string const & /* value */)
```
mkurdej: BTW, do you use clang-format?

quasardb/reader/ts_row.hpp
```diff
@@ -0,0 +1,250 @@
...
+ * columns eagerly, and as such will exhibit much worse performance.
+ */
+class ts_dict_row : public ts_row
+{
+public:
+    ts_dict_row()
+        : ts_row()
+    {}
+
+    ts_dict_row(qdb_local_table_t local_table, ts_columns_t const & columns)
```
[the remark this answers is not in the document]
solatis: Sorry! I figured clangformat would pick this up actually, I am conditioned by east const. West const it is, though. :)

quasardb/reader/ts_value.hpp
```diff
@@ -0,0 +1,184 @@
...
+    {}
+
+    ts_value(const ts_value & rhs)
+        : _local_table(rhs._local_table)
+        , _index(rhs._index)
+        , _type(rhs._type)
+    {}
+
+    ts_value(ts_value && rhs)
+        : _local_table(rhs._local_table)
```
mkurdej: I guess that members are just pointers and integers, so no move is necessary, but don't forget to `std::move` in other cases.

quasardb/reader/ts_row.hpp
solatis: Yes I use clang-format, it doesn't seem to pick up my const style fixes.

quasardb/reader/ts_value.hpp
```diff
@@ -0,0 +1,184 @@
...
+};
+
+template <typename Module>
+static inline void register_ts_value(Module & m)
+{
+    py::class_<ts_value>{m, "TimeSeriesReaderValue"};
+}
+
+}; // namespace reader
+}; // namespace qdb
```
mkurdej: `};` what are those semicolons?

quasardb/reader/ts_value.hpp
```diff
@@ -0,0 +1,184 @@
...
+     * C++->Python
+     */
+    static handle cast(qdb::reader::ts_value src, return_value_policy /* policy */, handle /* parent */)
+    {
+        return src.cast();
+    }
+};
+
+}; // namespace detail
+}; // namespace pybind11
```
mkurdej: Semicolons!

quasardb/ts.hpp
```diff
@@ -32,99 +32,114 @@
...
+        {
+            // This is a kludge, because technically a table can have no columns, and we're
+            // abusing it as "no argument provided". It's a highly exceptional use case, and
+            // doesn't really have any implication in practice, so it should be ok.
+            c_columns = list_columns();
+        }
+        else
+        {
+            c_columns.reserve(columns.size());
+            // This transformation can probably be optimized, but it's only invoked when cosntructing
```
mkurdej:
```suggestion
            // This transformation can probably be optimized, but it's only invoked when constructing
```

quasardb/ts.hpp
```diff
@@ -32,99 +32,114 @@
...
+            // the reader so it's unlikely to be a performance bottleneck.
+            for (auto a : columns)
```
mkurdej: Please rename `a`. `auto`?

You could do:
```
std::transform(std::cbegin(columns), std::cend(columns), std::back_inserter(c_columns), [this] (const auto & col) {
    return detail::column_info{this->column_type_by_id(col), col};
};
``

File: quasardb/ts.hpp [hunk 31]
```diff
@@ -32,99 +32,114 @@
...
+            {
+                c_columns.push_back(detail::column_info{this->column_type_by_id(a), a});
+            }
+        }
+
+        auto r = convert_ranges(ranges);
+
+        return (dict_mode == true
+                    ? py::cast(qdb::ts_reader<reader::ts_dict_row>(_handle, _alias, c_columns, r), py::return_value_policy::move)
+                    : py::cast(new qdb::ts_reader<reader::ts_fast_row>(_handle, _alias, c_columns, r), py::return_value_policy::move));
```
mkurdej: Why `new` on this line and not on the previous one???

quasardb/ts.hpp
```diff
@@ -222,6 +237,10 @@ class ts : public entry
 
         return res;
     }
+
+private:
+    bool _has_indexed_columns;
```
mkurdej: That's a bit nit-picking, but I'd rather make those members `mutable` and make "logically const" methods `const`.

quasardb/ts_convert.hpp
```diff
@@ -60,6 +60,7 @@ using time_ranges = std::vector<time_range>;
 
 static inline qdb_ts_range_t convert_range(const time_range & tr) noexcept
 {
+
```
mkurdej: That's called edouardosis, or sidneyosis. Or I-am-unable-to-set-up-my-editor-correctly-so-I-put-newline-at-the-beginning-of-each-file-function-or-block disease. :D

quasardb/ts_convert.hpp
```diff
@@ -72,6 +73,12 @@ static inline std::vector<qdb_ts_range_t> convert_ranges(const time_ranges & ran
     return res;
 }
 
+// TODO: Is the a more compliant way to describe 'all data in a table'?
+static inline time_ranges all_ranges()
+{
+    return time_ranges{{0, std::numeric_limits<std::int64_t>::max()}};
```
mkurdej: Well, there's `qdb_max_timespec` and `qdb_max_time` but Sidney reported to me recently that it doesn't work as expected, so your approach is OK.

mkurdej: And why `0`? Why not `std::numeric_limits<std::int64_t>::min()`? It's signed int.

quasardb/ts_reader.hpp
```diff
@@ -0,0 +1,189 @@
...
+        , _the_row(_local_table, _columns)
+    {}
+
+    ts_reader_iterator(qdb_local_table_t local_table, ts_columns_t const & columns)
+        : _local_table(local_table)
+        , _columns(columns)
+        , _the_row(_local_table, _columns)
+    {
+        // Immediately try to move to the first row
+        ++(*this);
```
mkurdej: So, if there are no rows, you throw on construction, right? That's intended? How does it behave?

quasardb/ts_reader.hpp
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    bool operator==(ts_reader_iterator const & rhs) const noexcept
```
mkurdej: That's not that important in this context, but I'd rather make them friend free functions `friend bool operator==(lhs, rhs) noexcept`,  so as to avoid possible implicit conversions.

start-services
```diff
@@ -1,19 +1,24 @@
 #!/bin/sh
 
+set -x
+
 CMD="qdb/bin/qdbd"
-ARGS="-a 127.0.0.1:28360 --security=false --storage-engine=transient --log-level=debug -r qdb/db"
-SECURE_ARGS="-a 127.0.0.1:28361 --security=true --cluster-private-file=cluster-secret-key.txt --user-list=users.txt --storage-engine=transient --log-level=debug -r qdb/securedb"
+ARGS="-c qdbd.conf"
```
mkurdej: I'd commit these changes (unrelated to bulk reader) in a separate commit.

users.txt
```diff
@@ -2,7 +2,9 @@
     "users": [
         {
             "username": "qdb-api-python",
+            "superuser": true,
```
mkurdej: This should be done in a separate commit (referring QDB-2381).

quasardb/ts_convert.hpp
Dwarfobserver: It's only for the 1st file line heretic !

mkurdej: [Uderz w stół, a nożyce się odezwą!](https://en.wiktionary.org/wiki/uderz_w_st%C3%B3%C5%82,_a_no%C5%BCyce_si%C4%99_odezw%C4%85)

quasardb/ts_convert.hpp
solatis: I figured negative time would be reserved for special values, or we would want that possibility in the future. Does it matter?

mkurdej: There is a special reserved time already (something like `timespec{min(), min()}`).

quasardb/ts_reader.hpp
solatis: No. Observe this code in operator++: if local table reached end, _local_table is set to nullptr, which should then match iterator end.

quasardb/ts_convert.hpp
mkurdej: Don't think it's important for now.

quasardb/ts_reader.hpp
solatis: Now that we talk about this, I'm more concerned about proper practice how to describe "end iterator" -- relying on `_local_table == nullptr` seems fragile to me. How is this typically handled?

quasardb/ts_reader.hpp
solatis: Anything that makes it less OOP gets my approval, will do. :)

quasardb/ts.hpp
solatis: Nice catch, both should be using move semantics.

quasardb/ts_reader.hpp
mkurdej: Ok. Well, a default-constructed (so `nullptr` here) iterator seems to me a common practice indeed.

quasardb/reader/ts_value.hpp
```diff
@@ -0,0 +1,184 @@
...
+        , _index(rhs._index)
+        , _type(rhs._type)
+    {}
+
+    ts_value(ts_value && rhs)
+        : _local_table(rhs._local_table)
+        , _index(rhs._index)
+        , _type(rhs._type)
+    {
+        rhs._local_table = nullptr;
```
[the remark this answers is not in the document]
solatis: Hmmm, I recall that my local_table was being released twice and this fixed it (setting _local_table to nullptr on rhs). I have to admit that I didn't fully dig into it.

quasardb/ts_convert.hpp
Dwarfobserver: 😮

quasardb/reader/ts_value.hpp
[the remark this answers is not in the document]
mkurdej: Hmmm, so `ts_value` destroys its local table? Shouldn't it be some sort of shared pointer then? Is `ts_value` movable-only then?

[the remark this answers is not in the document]
solatis: Oh crap this is ts_value not ts_reader. Ignore my remark, will fix.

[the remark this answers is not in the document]
solatis: ts_value is supposed to be a really thin wrapper and not hold/own any resources of itself.

quasardb/reader/ts_value.hpp
solatis: The antichrist?

quasardb/ts_convert.hpp
edouarda: I like to have space in my text :-)

quasardb/ts_reader.hpp
```diff
@@ -0,0 +1,189 @@
...
+    ts_reader_iterator(qdb_local_table_t local_table, const ts_columns_t & columns)
+        : _local_table(local_table)
+        , _columns(columns)
+        , _the_row(_local_table, _columns)
+    {
+        // Immediately try to move to the first row
+        ++(*this);
+    }
+
+    bool operator==(const ts_reader_iterator & rhs) const noexcept
```
edouarda: noexcept?

quasardb/ts_reader.hpp
```diff
@@ -0,0 +1,189 @@
...
+
+    iterator end()
+    {
+        return iterator();
+    }
+
+private:
+    qdb::handle_ptr _handle;
+    ts_columns_t _columns;
+    qdb_local_table_t _local_table;
```
edouarda: you can also do qdb_local_table_t _local_table{nullptr};

quasardb/detail/ts_column.hpp
```diff
@@ -0,0 +1,144 @@
...
+    column_info(const qdb_ts_column_info_t & ci)
+        : column_info{ci.type, ci.name}
+    {}
+
+    operator qdb_ts_column_info_t() const noexcept
+    {
+        qdb_ts_column_info_t res;
+
+        res.type = type;
+        res.name = name.c_str();
```
mkurdej: Hmm, you don't need a deep copy here?

quasardb/reader/ts_row.hpp
```diff
@@ -0,0 +1,247 @@
...
+{
+
+using ts_columns_t = std::vector<detail::column_info>;
+
+class ts_row
+{
+public:
+    // We need a default constructor to due being copied as part of an iterator.
+    ts_row() noexcept
+        : _local_table{nullptr}
```
mkurdej: You could have done `qdb_local_table_t _local_table{nullptr};` at member declaration and just do `= default;` here.

quasardb/reader/ts_row.hpp
```diff
@@ -0,0 +1,247 @@
...
+        // not implemented
+    }
+
+    std::string repr() const
+    {
+        // This is inefficient because we first cast everything to Python objects, get the
+        // string representation and then cast back to c++ strings, but it's correct and
+        // effective, and it's used for debugging purposes only.
+        auto xs         = copy();
+        std::string kvs = std::accumulate(xs.cbegin(), xs.cend(), std::string(), [](std::string & acc, auto x) {
```
mkurdej:
```suggestion
        std::string kvs = std::accumulate(xs.cbegin(), xs.cend(), std::string(), [](std::string && acc, auto x) {
```
Use rvalue and move below.

quasardb/reader/ts_row.hpp
```diff
@@ -0,0 +1,247 @@
...
+            std::string s = py::str(x);
+            return acc.empty() ? s : acc + ", " + s;
```
mkurdej:
```suggestion
            return acc.empty() ? s : std::move(acc) + ", " + std::move(s);
```

quasardb/reader/ts_row.hpp
```diff
@@ -0,0 +1,247 @@
...
+        });
+
+        return "[" + kvs + "]";
```
mkurdej:
```suggestion
        return '[' + std::move(kvs) + ']';
```

quasardb/reader/ts_row.hpp
```diff
@@ -0,0 +1,247 @@
...
+    std::string repr() const
+    {
+        // Same as ts_fast_row::repr, this is inefficient but it's ok
+        auto xs         = copy();
+        std::string kvs = std::accumulate(xs.cbegin(), xs.cend(), std::string(), [](const std::string & acc, const auto & x) {
+            std::string s = "'" + x.first + "': " + std::string(py::str(x.second));
+            return acc.empty() ? s : acc + ", " + s;
+        });
+
+        return "{" + kvs + "}";
```
mkurdej:
```suggestion
        return '{' + std::move(kvs) + '}';
```

quasardb/reader/ts_row.hpp
```diff
@@ -0,0 +1,247 @@
...
+    constexpr void set_item(const std::string & /* alias */, const std::string & /* value */) const noexcept
+    {
+        // not implemented
+    }
+
+    std::string repr() const
+    {
+        // Same as ts_fast_row::repr, this is inefficient but it's ok
+        auto xs         = copy();
+        std::string kvs = std::accumulate(xs.cbegin(), xs.cend(), std::string(), [](const std::string & acc, const auto & x) {
```
mkurdej:
```suggestion
        std::string kvs = std::accumulate(xs.cbegin(), xs.cend(), std::string(), [](std::string && acc, const auto & x) {
```

quasardb/reader/ts_row.hpp
```diff
@@ -0,0 +1,247 @@
...
+            std::string s = "'" + x.first + "': " + std::string(py::str(x.second));
+            return acc.empty() ? s : acc + ", " + s;
```
mkurdej:
```suggestion
            return acc.empty() ? s : std::move(acc) + ", " + std::move(s);
```

quasardb/reader/ts_row.hpp
mkurdej: If `string &&` doesn't work (e.g. with older STL implementations), just pass by value.

tests/test_ts_reader.py
```diff
@@ -5,6 +5,9 @@
 import test_ts_batch as batchlib
 import numpy as np
 
+def test_reader_can_return_no_rows(qdbd_connection, table, many_intervals):
+    assert 0 == reduce(lambda x,y : x+1, table.reader(), 0)
```
mkurdej: Don't forget to run auto formatter :) `autopep8`.

quasardb/ts_reader.hpp
```diff
@@ -58,7 +58,7 @@ class ts_reader_iterator
     using value_type = RowType;
     using reference  = const value_type &;
 
-    friend bool operator==<>(const ts_reader_iterator<RowType> &, const ts_reader_iterator<RowType> &);
+    friend bool operator==<>(const ts_reader_iterator<RowType> &, const ts_reader_iterator<RowType> &) noexcept;
```
mkurdej: ?????? `operator==<>`

solatis: It's a templated operator==, of course!

quasardb/detail/ts_column.hpp
solatis: Funny, this was already here but I think you're right. The reason this works is that all the std::string objects stay alive longer than these column_info objects (it's used for doing python -> qdb stuff, for example when creating a table). 

It's fragile. I'm inclined to fix this, but it would be a performance regression. What are your thoughts?

mkurdej: Let it as is for the moment. Add a comment in the code though.

PR #12, opened 2019-05-10
[QDB-2491] Add integer API

solatis:


.gitignore
```diff
@@ -3,7 +3,6 @@
 /bin/
 /build*
 /project/
-/quasardb/
```
vianneyPL: Are you sure about this?
Maybe ignore any file in it, and put a `.gitkeep` or something alike in the directory?

solatis: Why ? This is the actual API C++ code in that path

solatis: Marek added that in this patch

https://github.com/bureau14/qdb-api-python/commit/f6ec5d7b3fce01450a5a95fba9621e0440842e54

But I think it was a mistake.

PR #19, opened 2020-01-29
QDB-2547 - Add client-api logging ability to Python API

solatis:
Adds logging capabilities to the Python API. Most interesting to review is mostly the logger.cpp / logger.hpp.

### DIFF
--- a/quasardb/error.hpp
+++ b/quasardb/error.hpp
@@ -126,6 +126,12 @@ template <typename PreThrowFtor = detail::no_op>
 void qdb_throw_if_error(qdb_error_t err, PreThrowFtor && pre_throw = detail::no_op{})
 {
     static_assert(noexcept(std::forward<PreThrowFtor &&>(pre_throw)()), "`pre_throw` argument must be noexcept");
+
+
+    // HACKS(leon): we need to flush our log buffer a lot, ideally after every native qdb
+    //              call. Guess which function is invoked exactly at those moments?
+    qdb::native::flush();
+
     if ((qdb_e_ok != err) && (qdb_e_ok_created != err))
     {
       pre_throw();
@@ -141,10 +147,6 @@ void qdb_throw_if_error(qdb_error_t err, PreThrowFtor && pre_throw = detail::no_
         throw qdb::exception{err};
       };
     }
-
-    // HACKS(leon): we need to flush our log buffer a lot, ideally after every native qdb
-    //              call. Guess which function is invoked exactly at those moments?
-    qdb::native::flush();
 }
 
 template <typename Module>
--- a/quasardb/error.hpp
+++ b/quasardb/error.hpp
@@ -30,6 +30,7 @@
  */
 #pragma once
 
+#include "logger.hpp"
 #include <qdb/error.h>
 #include <utility>
 #include <pybind11/pybind11.h>
@@ -140,6 +141,10 @@ void qdb_throw_if_error(qdb_error_t err, PreThrowFtor && pre_throw = detail::no_
         throw qdb::exception{err};
       };
     }
+
+    // HACKS(leon): we need to flush our log buffer a lot, ideally after every native qdb
+    //              call. Guess which function is invoked exactly at those moments?
+    qdb::native::flush();
 }
 
 template <typename Module>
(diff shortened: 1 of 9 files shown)


quasardb/pandas/__init__.py
```diff
@@ -253,11 +275,18 @@ def write_dataframe(df, cluster, table, create=False, _async=False, chunk_size=5
                     fn = write_with[ct]
                     fn(i, v)
 
+        start = time.time()
+
+        logger.debug("push chunk of %d rows, fast?=%s, async?=%s", len(current_df.index), fast, _async)
+
+        if fast is True:
+            batch.push_async()
```
mkurdej: Didn't you mean:
```suggestion
            batch.push_fast()
```
?

quasardb/pandas/__init__.py
```diff
@@ -175,6 +186,7 @@ def read_dataframe(table, row_index=False, columns=None, ranges=None):
         columns = list(c.name for c in table.list_columns())
 
     if not row_index:
+        logger.debug("reading DataFrame from %d series", len(columns))
```
mkurdej: series? or columns?

quasardb/logger.hpp
```diff
@@ -0,0 +1,223 @@
...
+   */
+  logger(const std::string & module_name)
+  {
+
+    // TODO(leon): is there a way we can (safely) cache these calls? they should
+    //             remain the same for every python VM instance... is it safe to cache?
+    //
+    // NOTE(leon): looking at the pybind11 code, it appears they never cache anything.. so
+    //             that might hint that it's complex to get right
+    py::module logging = py::module::import("logging");
```
mkurdej: Wouldn't it work to make them, hmmm, function static variables?
```suggestion
    static py::module logging = py::module::import("logging");
```

quasardb/error.hpp
```diff
@@ -140,6 +141,10 @@ void qdb_throw_if_error(qdb_error_t err, PreThrowFtor && pre_throw = detail::no_
         throw qdb::exception{err};
       };
     }
+
+    // HACKS(leon): we need to flush our log buffer a lot, ideally after every native qdb
+    //              call. Guess which function is invoked exactly at those moments?
+    qdb::native::flush();
```
mkurdej: Why don't you flush when there is an error? I'd put it before the `if`. Or am I missing something?

quasardb/logger.cpp
```diff
@@ -0,0 +1,108 @@
...
+    qdb::native::_do_flush();
+  }
+}
+
+
+void
+qdb::native::_do_flush() {
+  // NOTE: this assumes a lock to _buffer-lock has been acquired, never call this
+  //       function directly!
+
```
mkurdej: `assert(!_buffer_lock.try_lock());`
hmm?

quasardb/logger.hpp
```diff
@@ -0,0 +1,223 @@
...
+   * As such, try to keep this logging out of the critical performance path.
+   */
+class logger
+{
+public:
+
+  /**
+   * Default constructor, invoking any .debug() etc on a default initialized
+   * instantiation of this class leads to undefined behavior, but it's needed
+   * for certain situations.
```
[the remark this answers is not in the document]
solatis: Actually after re-reading the code a few times, the lock in `swap_callback` actually ensures this isn't possible. But basically QuasarDB would need to start sending callbacks _before_ the _logger is initialized.

quasardb/logger.hpp
solatis: Unfortunately, all these Python objects work with a reference to the 'Python Interpreter' under the hood. I don't think there is a strict guarantee that this reference will always be valid, only in the context of a function call.

quasardb/error.hpp
solatis: Good point.

quasardb/pandas/__init__.py
solatis: Yeah as this is somewhat user-visible, and relates to "Building a Pandas dataframe out of a number of Pandas series", I figured it was best to stick with Pandas terminology.

solatis: But a DataFrame == a table, and a TimeSeries == a column per pandas <> qdb terminology, for what it's worth.

PR #21, opened 2020-05-20
Pin numpy < 1.19 to avoid problem with 1.19.0rc1 failing build on Linux, Python 3.5.

mkurdej:
Is that the course of action we should take? Is there another solution? How about banning RC versions of dependencies?


solatis:
This patch would require all our users to use this exact same version of numpy, which is suboptimal. I think the best course of action right now would be to explicitly say < 1.17.4, until we implement compatibility for this version.

mkurdej:
How about now?

solatis:
Good to go! :+1:

PR #29, opened 2021-02-08
Fix crash on exit

solatis:


quasardb/logger.hpp
```diff
@@ -132,31 +117,49 @@ class logger
     template <typename... Args>
     void critical(Args &&... args) const
     {
-        _log(_critical, std::forward<Args>(args)...);
+        _log(_module_name,
+             "critical",
+             std::forward<Args>(args)...);
```
mkurdej, line 122: You could add a helper `_log(level, msg, args)` function that will call `_log(_module_name, ...)`.

mkurdej (approved):
Apart from the _log comment. FWIW, LGTM. :)

PR #30, opened 2021-06-02
QDB-9669 - Build with centos7.

vianneyPL:


scripts/teamcity/10.build.sh
```diff
@@ -1,6 +1,16 @@
+#!/bin/bash
+
+# Using cento7 docker we need to enable gcc-8 and g++-8
```
mkurdej, line 3:
```suggestion
# Using centos7 docker we need to enable gcc-8 and g++-8
```

scripts/teamcity/20.test.sh
```diff
@@ -1,5 +1,16 @@
+#!/bin/bash
+
 PYTHON="${PYTHON_CMD:-python}"
 
+# Using cento7 docker we need to enable gcc-8 and g++-8
```
mkurdej, line 5:
```suggestion
# Using centos7 docker we need to enable gcc-8 and g++-8
```

scripts/teamcity/20.test.sh
```diff
@@ -1,5 +1,16 @@
...
+case "$(uname)" in
+    Linux*)
+        source /opt/rh/devtoolset-8/enable
```
mkurdej, line 8: Isn't it enough to do it once only in `10.build.sh`?

scripts/teamcity/20.test.sh
```diff
@@ -29,6 +40,7 @@ esac
 # first remove system then user
 ${PYTHON} -m pip uninstall -r dev-requirements.txt -y
 ${PYTHON} -m pip uninstall -r dev-requirements.txt -y
```
mkurdej, line 42: But... that's the same command twice. Why?

docker/Dockerfile
```diff
@@ -1,7 +1,46 @@
...
+FROM centos:7
 
-RUN apt-get update \
-        && apt-get install -y cmake \
-        && apt-get clean
+ARG PYTHON_VERSION
+
+RUN yum install -y \
+
+        # For recent GCC
```
mkurdej, line 9: I find it error-prone to have blank lines and comments in multi-lines (with trailing `\`).

scripts/teamcity/20.test.sh
vianneyPL: It is not. It only replaces the current shell.

I could try to put it in the bashrc or something like that.

scripts/teamcity/20.test.sh
vianneyPL: First one remove from the local user. Second one remove system-wide.

mkurdej: But... no need to `sudo` or whatever to remove system-wide?

PR #34, opened 2021-10-04
QDB-9794 - Set batch options

vikonix:


mkurdej (changes requested):
Add a test.

mkurdej (approved):
My fault. I thought this change was in the qdb repo which would need a test of course. But here we already have failing tests.
Also, there's no way we can default-initialize batch.options member automatically in C API.
LGTM.

PR #35, opened 2021-11-12
QDB-9711 - Allow shard size specification in  pandas write_dataframe.

vianneyPL:


solatis (approved):
Nothing to comment, looks good to me!

PR #36, opened 2022-06-17
Fix datetime <> chrono <> qdb_time_t conversion for timezones

solatis:
Fixes timezone / expiry issue, and overall fixes incorrect assumptions about system_clock / local_time / datetime.datetime:

 - Implements custom type wrappers for Python `datetime` objects:
   - `datetime.datetime`
   - `datetime.tzinfo`
   - `datetime.timedelta`
 - Implements conversion routines to/from datetime and chrono time_points:
   - if the datetime object has an associated timezone, converts into a chrono::system_clock::time_point that represents the UTC time;
   - if the datetime object does *not* have an associated timezone, assumes it is in local time.

Python's datetime is a bit messy, as a timezone is optional, and not required. Specifically, both `datetime.now()` and `datetime.utcnow()` created datetime objects _without_ time zone, one being in localtime and the other in UTC.

The use of `datetime.utcnow()` has been discourage by Python official, in favor of `datetime.now(tz=timezone.utc)`, which *does* produce a datetime object with timezone. As such, the best we can do is treat these "naked" datetime objects as local time.

A tradeoff I made was not to include and vendor the `date` thirdparty library, which was only necessary for the `year_month_day` and `hh_mm_ss` objects. Instead, we'll use these same classes from C++20. This required updating GCC to 11 in the Linux builds, which still produces `manylinux2014` compatible packages compatible with glibc >= 2.17.


quasardb/pytypes.hpp
```diff
@@ -0,0 +1,221 @@
...
+        return tz_;
+    };
+
+    inline pytimedelta utcoffset() const noexcept
+    {
+        return tzinfo().utcoffset(*this);
+    };
+};
+
+}; // namespace qdb
```
mkurdej, line 221:
```suggestion
} // namespace qdb
```
Nit  :)

quasardb/pytypes.hpp
```diff
@@ -0,0 +1,221 @@
...
+            // either `datetime.now()` or `datetime.utcnow()`.
+            //
+            // since the use of `datetime .utcnow()` is actively discouraged in favor
+            // of `datetime.now(tz=timezone.utc)`, we are going to assume that no timezone
+            // means local time.
+            //
+            // the most elegant way to handle this is to amend this datetime
+            // object with the local timezone and recurse.
+            return astimezone().tzinfo();
+        };
```
mkurdej, line 210:
```suggestion
        }
```

quasardb/pytypes.hpp
```diff
@@ -0,0 +1,221 @@
...
+
+    inline int second() const noexcept
+    {
+        return PyDateTime_DATE_GET_SECOND(ptr());
+    };
+
+    inline int microsecond() const noexcept
+    {
+        return PyDateTime_DATE_GET_MICROSECOND(ptr());
+    };
```
mkurdej, line 158:
```suggestion
    }
```
Here and elsewhere.

quasardb/masked_array.hpp
```diff
@@ -35,7 +35,6 @@
 #include "traits.hpp"
 #include <pybind11/numpy.h>
 #include <pybind11/stl_bind.h>
-#include <iostream>
```
mkurdej, line 38: Amen!

quasardb/blob.hpp
```diff
@@ -98,29 +97,25 @@ class blob_entry : public expirable_entry
         return convert_and_release_content(content, content_length);
     }
 
-    pybind11::bytes get_and_update(const std::string & data,
-        std::chrono::system_clock::time_point expiry = std::chrono::system_clock::time_point{})
+    pybind11::bytes get_and_update(const std::string & data)
```
mkurdej, line 100: I gather that we never used nor tested the expiry.

mkurdej (approved):
Nice!

quasardb/blob.hpp
solatis: Yes, this code was entirely untested and unused, and is mostly a convenience; you can still set the expiry by using `expires_at()`, so I figured it's better to just simplify and have only one code path.

quasardb/pytypes.hpp
solatis: Is there a way to tell clang-format that I want this ?

mkurdej: WDYM? It already formats it correctly.
Or you just called me "clang-format"? :D

solatis: All this code was already ran through clang-format, so if it's incorrect, wouldn't that imply the formatting rules are insufficient?

mkurdej: It's not incorrect. `;` is an empty statement. You can put as many of them as you want.
That was just a stylistic remark.
One puts `;` only after a struct's/class's closing brace.

mkurdej: Yeah, it doesn't remove semicolons (yet).
It may however (since recently) add or remove braces. Semicolons will come one day ;).

PR #38, opened 2022-11-18
Add support for AArch64

mkurdej:
AArch64 builds need to pass `--build-arg ARCH="aarch64"` and `--build-arg DEVTOOLSET_VERSION=10` when building Docker image.


docker/Dockerfile
```diff
@@ -4,15 +4,16 @@ ARG PLATFORM="${POLICY}_${ARCH}"
 
 FROM quay.io/pypa/$PLATFORM
 
-ENV PATH="/opt/rh/devtoolset-11/root/usr/bin:$PATH"
+ARG DEVTOOLSET_VERSION="11"
```
solatis, line 7: ARM needs a different one?

quasardb/convert/value.hpp
```diff
@@ -37,6 +37,7 @@
 #include "../traits.hpp"
 #include "unicode.hpp"
 #include <qdb/ts.h>
+#include <date/date.h> // We cannot use <chrono> until we upgrade to at least GCC11 (ARM).
```
solatis, line 40: How do the APIs of date/chrono compare? Looks like they're roughly equivalent?

quasardb/convert/value.hpp
```diff
@@ -364,8 +366,8 @@ struct value_converter<traits::bytestring_dtype, qdb_string_t>
     using char_t = std::string::value_type;
 
     template <concepts::input_range_t<char_t> R>
-    requires(ranges::sized_range<R> && ranges::contiguous_range<R>) inline qdb_string_t operator()(
-        R && x) const
+        requires(ranges::sized_range<R> && ranges::contiguous_range<R>)
+    inline qdb_string_t operator()(R && x) const
```
solatis, line 370: Are you using  a different clang-format / linting config? If so, can you somehow sync it into this project?

As far as I'm concerned all these files were already linted using clang-format and these changes are a bit of noise :)

docker/Dockerfile
solatis: Did you already add a teamcity project for this?

solatis (commented):
Looks solid, you had no issues with the concepts / ranges / modern c++ stuff with ARM?

docker/Dockerfile
mkurdej: Yes, https://teamcity.quasardb.net/buildConfiguration/Api_Python_Release_Linux_Python37AArch64?branch=test&buildTypeTab=overview&mode=builds.

mkurdej: amazonlinux image has no GCC 11, so I used 10. This in turn, doesn't have date parts in `<chrono>`, hence I added date library.

quasardb/convert/value.hpp
mkurdej: They are pretty much identical. Standard is based on date library.

quasardb/convert/value.hpp
mkurdej: I think that it's rather clang-format version that differs. Earlier versions weren't able to format requires clauses.

mkurdej: I'll do formatting changes separately.

solatis: Ah makes sense then!

mkurdej:
> Looks solid, you had no issues with the concepts / ranges / modern c++ stuff with ARM?

No problem at all. GCC 10 is enough.
Any other remarks?

solatis:
No other comments

solatis (approved):

PR #39, opened 2023-03-24
QDB-10608 - Python API should allow secure connections by providing cluster public / user private key files directly

vikonix:


solatis (commented):
Other than that we should probably add support for this for direct connect nodes as well (node.hpp), the changes + tests look good to me.

PR #60, opened 2024-07-17
QDB-14207 - Fix Python error in ODBC driver tests

gbfz:
the ODBC driver tests [fail on windows teamcity agents](http://teamcity-server:8111/buildConfiguration/Tools_OdbcDriver_Win32/411380?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildDeploymentsSection=false&expandBuildProblemsSection=true&showLog=411320_8222_1333.8222&logFilter=debug&logView=flowAware) because of a text encoding issue.
the same build [passes on linux](http://teamcity-server:8111/buildConfiguration/Tools_OdbcDriver_Linux/411393?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildDeploymentsSection=false), suggesting that utf-8 is preferred for tests to work normally.
I added a step in the api initialization to enforce utf-8 usage.

### DIFF
--- a/quasardb/__init__.py
+++ b/quasardb/__init__.py
@@ -119,9 +119,5 @@ if not 'quasardb' in locals():
     raise ImportError()
 
 
-import sys
-sys.stdout.reconfigure(encoding='utf-8')
-sys.stderr.reconfigure(encoding='utf-8')
-
 from .extensions import extend_module
 extend_module(quasardb)
--- a/quasardb/__init__.py
+++ b/quasardb/__init__.py
@@ -119,5 +119,9 @@ if not 'quasardb' in locals():
     raise ImportError()
 
 
+import sys
+sys.stdout.reconfigure(encoding='utf-8')
+sys.stderr.reconfigure(encoding='utf-8')
+
 from .extensions import extend_module
 extend_module(quasardb)
--- a/tests/test_pandas.py
+++ b/tests/test_pandas.py
@@ -119,8 +119,6 @@ def test_dataframe(qdbpd_write_fn, df_with_table, qdbd_connection):
 
     df2 = qdbpd.read_dataframe(table)
 
-    print("df2: {}".format(df2))
-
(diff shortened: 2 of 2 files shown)


vikonix (approved):
lgfm

solatis:
Do we know what change actually triggered this issue? Are we sure we're not just fixing a symptom, rather than the root cause?

vikonix:
What is the Windows localization we use for ODBC testing?
Pretty sure that we use not US.
The first build failed 12.06 without any sources was changed.

gbfz:
the win32 build has been failing since 12.06 with no changes, but the exact reasons for failure changed overtime from parallelism test failure to a setup error to the encoding error.
my branch is green, so it fixed the encoding issue, but I cannot explain what fixed other failures

solatis:
I just realized that this patch was made in the actual python API, rather than the ODBC driver. This is not an acceptable fix, we need to find the root cause what caused this problem. We cannot just change the behavior of all applications that use the quasardb Python API, we need to figure out the root cause what caused this problem and fix that.

solatis:
Also please note that none of the tests in the Python API have been failing, so I’m inclined to think it’s related to the ODBC driver implementation instead.

vikonix:
http://teamcity-server:8111/buildConfiguration/Api_Python_Debug_Windows_Win32_Python310/404438?expandBuildDeploymentsSection=false&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true&showLog=404438_7647_7647&logFilter=debug&logView=flowAware
we have fully same problem on the pure Windows python API tests

vikonix:
http://teamcity-server:8111/project/Api_Python_Debug_Windows_Win32?mode=builds
you can see here that George path fixed this tests

solatis (approved):

PR #62, opened 2024-07-22
QDB-13421 -  Add support for the new compression flag in the Python API

gbfz:


tests/test_options.py
```diff
@@ -30,9 +28,34 @@ def test_set_client_max_parallelism_8(qdbd_settings):
...
+def test_compression_none(qdbd_connection):
+    qdbd_connection.options().set_compression(
+        quasardb.Options.Compression.Disabled)
+
+def test_compression_balanced(qdbd_connection):
+    qdbd_connection.options().set_compression(quasardb.Options.Compression.Balanced)
+
+@pytest.mark.skip(reason="Not yet supported")
+def test_compression_best(qdbd_connection):
+    qdbd_connection.options().set_compression(quasardb.Options.Compression.Best)
```
solatis, line 53: Perhaps it's better to instead verify that an exception is thrown? Now you're just not executing the test, it strikes me that we may want to test that the flag exists but doesn't work (yet).

tests/test_options.py
```diff
@@ -30,9 +28,34 @@ def test_set_client_max_parallelism_8(qdbd_settings):
...
+
+@pytest.mark.skip(reason="Not yet supported")
+def test_compression_fast(qdbd_connection):
+    qdbd_connection.options().set_compression(quasardb.Options.Compression.Fast)
```
solatis, line 57: Same for this as well, probably want to verify an exception is thrown, i.e. `with pytest.raises(quasardb.Error)` or something along those lines.

solatis (changes requested):

solatis (approved):

PR #67, opened 2024-08-08
QDB-14333 -  Ensure write_through options are exposed in the Python API

gbfz:


quasardb/numpy/__init__.py
```diff
@@ -652,6 +653,11 @@ def write_arrays(
 
       Defaults to False.
 
+    write_through: optional bool
+      Tell the server to not cache data after write.  This flag can have a positive
+      or negative impact on performance and should be used carefully,
+      by default caching is left at the discretion of the server.
```
solatis, line 659: Sounds a bit too alarmist. 

I would phrase it as:

"If True, data is not cached after write. By default is False, in which case caching is left at the discretion of the server".

quasardb/pandas/__init__.py
```diff
@@ -378,6 +379,11 @@ def write_dataframes(
 
       Defaults to False.
 
+    write_through: optional bool
+      Tell the server to not cache data after write.  This flag can have a positive
+      or negative impact on performance and should be used carefully,
+      by default caching is left at the discretion of the server.
+
     """
```
solatis, line 387: Same comment as above.

quasardb/writer.cpp
```diff
@@ -392,6 +412,27 @@ detail::deduplicate_options writer::_deduplicate_from_args(py::kwargs args)
...
+        return py::cast<bool>(args["write_through"]) ? qdb_exp_batch_push_flag_write_through
+                                                     : qdb_exp_batch_push_flag_none;
+    }
+    else
+    {
+        std::string error_msg = "Invalid argument provided for `write_through`: expected bool, got: ";
+        error_msg += args["write_through"].cast<py::str>();
+
+        throw qdb::invalid_argument_exception{error_msg};
+    }
```
solatis, line 433: You'll want to print the type here instead, rather than the value. It currently can be confusing, e.g. if someone provided the literal string "True".

quasardb/writer.cpp
```diff
@@ -392,6 +412,27 @@ detail::deduplicate_options writer::_deduplicate_from_args(py::kwargs args)
...
+    if (!args.contains("write_through"))
+    {
+        return qdb_exp_batch_push_flag_none;
+    }
+
+    if (py::isinstance<py::bool_>(args["write_through"]))
+    {
+        return py::cast<bool>(args["write_through"]) ? qdb_exp_batch_push_flag_write_through
+                                                     : qdb_exp_batch_push_flag_none;
+    }
```
solatis, line 426: I think it may be more efficient to *assume* it's True (i.e. immediately do the cast), and catch a py::TypeError (or whatever the exception is);

The reason is that `isinstance()` happens to be a super expensive function call, and right now you're basically doing two different casts. 

The default case is that it is absolutely a bool, and it would only be the case when someone's developing new code that they may accidentally provide the wrong type. As such, we can safely assume that it will be a bool, and only in highly exceptional cases not.

solatis (approved):
Great work!

I pre-approved this PR, but left a few comments that require some small changes. Feel free to merge once addressed, and don't forget to merge into 3.14.x as well (you should be able to do this freely now, as I have removed the branch protection).

PR #70, opened 2024-08-14
QDB-14429 - Add test cases for retries

solatis:
This adds test cases for retries. 

What it does:
 * optionally mocks failures;
 * verifies the number of retries seen in the logs;
 * verifies the time slept in the logs.

There's also a small refactoring of the arguments accepted by the Pandas adapter to simplify things and pass arguments as-is to the numpy adapter.


quasardb/detail/retry.hpp
```diff
@@ -39,70 +40,147 @@ namespace qdb::detail
...
+    }
+};
+
+// A mock when running tests, to trigger 'please retry' errors for this amount of times,
+// only enabled when building for tests.
+//
+// We take extra precautions that parts of the code are only compiled when tests are built,
+// but for convenience we do always expose this class type and its variables, to avoid preprocessor
+// ifdef/ifndef madness all over the place and keep function signatures simple.
+struct mock_failure_options
```
gbfz, line 137: is this actually needed? afaict the actual usage of `mock_failure_options` is identical to `retry_options` — the `err` field is never set to anything but `qdb_e_async_pipe_full`

quasardb/writer.cpp
```diff
@@ -503,12 +526,27 @@ qdb_uint_t writer::_push_flags_from_args(py::kwargs args)
...
 }
 
-void writer::_do_push(qdb_exp_batch_options_t const & options,
-    std::vector<qdb_exp_batch_push_table_t> const & batch,
-    detail::retry_options const & retry_options)
+void writer::_do_push(                                     //
+    qdb_exp_batch_options_t const & options,               //
+    std::vector<qdb_exp_batch_push_table_t> const & batch, //
+    detail::retry_options retry_options,                   //
+    detail::mock_failure_options mock_failure_options      //
```
gbfz, line 533: we could have the `mock_failure_options` argument be of type `retry_options` since their usage is identical

PR #86, opened 2025-04-01
QDB-16270 - Deprecate push modes in qdbnp.write_dataframes

rodp63:


quasardb/detail/writer.hpp
```diff
@@ -480,6 +482,23 @@ struct batch_push_mode
...
+	    return qdb_exp_batch_push_fast;
+	} else if (push_mode == "async") {
+	    return qdb_exp_batch_push_async;
+	} else if (push_mode == "truncate") {
+	    return qdb_exp_batch_push_truncate;
+	} else {
+	    throw qdb::invalid_argument_exception{"Invalid push mode"};
+	}
+    }
 };
```
solatis, line 502: Is this executed a lot? This is not fast code, as a matter of fact, it's super slow. Just something to be aware of.

quasardb/numpy/__init__.py
```diff
@@ -705,20 +708,35 @@ def write_arrays(
...
+      The mode used for inserting data. Can be either a string or a `WriterPushMode` enumeration item.
+      Available options:
+      * `'truncate'`: Truncate (also referred to as upsert) the data in-place. Will detect time range
+        to truncate from the time range inside the dataframe.
+      * `'async'`: Uses asynchronous insertion API where commits are buffered server-side and
+        acknowledged before they are written to disk. If you insert to the same table from
+        multiple processes, setting this to True may improve performance.
+      * `'fast'`: Whether to use 'fast push'. If you incrementally add small batches of data to table,
+        you may see better performance if you set this to True.
+      * `'transactional'`: Ensures full transactional consistency.
```
solatis, line 721: Why not directly use the `quasardb.WriterPushMode.Truncate` etc enum and make use of pybind11's automatic conversion? Why use a string here when we have this enum type already?

quasardb/numpy/__init__.py
```diff
@@ -838,20 +859,67 @@ def write_arrays(
...
+                "Please explicitly set push_mode='truncate'.",
+                DeprecationWarning,
+                stacklevel=deprecation_stacklevel,
+            )
+            push_kwargs["range"] = truncate
+            push_kwargs["push_mode"] = quasardb.WriterPushMode.Truncate
+
+        elif _async is True:
+            deprecation_warning("_async", "async")
+            push_kwargs["push_mode"] = quasardb.WriterPushMode.Async
```
solatis, line 910: This code can be done much more elegantly by using recursion.

Get rid of the check for `sum()` at the top.

I would then do this:

```
kwargs_to_push_mode = {'_async': quasardb.WriterPushMode.Async,
                                              'fast': quasardb.WriterPushMode.Fast,
                                             ...}
                                             
for k,v in kwargs_to_push_mode.items():
   if k in kwargs:
      if push_mode is not None:
          raise InvalidArgumentError("found '{}' in keywords, but push mode is already set to: {}".format(k, push_mode)
       kwargs['push_mode'] = v
       return         return write_arrays(data, cluster, table=table,
            dtype=dtype,
            index=index,
            _async=_async,
            fast=fast,
            truncate=truncate,
            deduplicate=deduplicate,
            deduplication_mode=deduplication_mode,
            infer_types=infer_types,
            write_through=write_through,
            writer=writer,
            retries=retries,
            **kwargs,
        )
```

The recursion can probably be handled more elegantly, but I'm seeing too many if-else statements here that can be replaced by a simple lookup map/dict + recursion. I

quasardb/writer.hpp
```diff
@@ -328,7 +328,8 @@ static void register_writer(py::module_ & m)
         .value("Transactional", qdb_exp_batch_push_transactional)                            //
         .value("Fast", qdb_exp_batch_push_fast)                                              //
         .value("Truncate", qdb_exp_batch_push_truncate)                                      //
-        .value("Async", qdb_exp_batch_push_async);                                           //
+        .value("Async", qdb_exp_batch_push_async)                                            //
+        .def_static("from_string", &detail::batch_push_mode::from_string);                   //
```
solatis, line 332: I don't like this, exposing this static c++ function like this, it's too and more of an internal C++ utility function. I don't think this function is that useful if we just use the enum type we already have.

solatis (changes requested):
Two major feedbacks:
 * Don't use another "string" representation for the push mode, just use the actual enum types directly;
 * I don't think the `from_string` function is necessary anymore after we do this, this should simplify the python code;
 * Instead of a whole bunch of if/elif statements, use a lookup map `kw_to_push_mode`, iterate over that, and (potentially) use recursion.

I typically like recursion for situations like this, but the function argument parsing may need some refactoring for that to work nicely here without too much copy/paste.

quasardb/detail/writer.hpp
rodp63: I agree on not using string conversion, let's use directly the enumeration

quasardb/numpy/__init__.py
rodp63: Now we're using a lookup rather than direct `if / else` statements. All the cases are handled, specially the `truncate` one, since it allowed the pass of a truncation range.

rodp63:
Deprecation warnings and kwargs combinations:
```python
qdbnp.write_arrays(data, conn, index=idx, fast=True)
# DeprecationWarning: The argument 'fast' is deprecated and will be removed in a future version. Please use 'push_mode=WriterPushMode.Fast' instead.
```

```python
qdbnp.write_arrays(data, conn, index=idx, truncate=(0,0))
# DeprecationWarning: The argument 'truncate' is deprecated and will be removed in a future version. Please use 'push_mode=WriterPushMode.Truncate, truncate_range=<RANGE>' instead.
```

```python
qdbnp.write_arrays(data, conn, index=idx, fast=True, push_mode=quasardb.WriterPushMode.Truncate)
# InvalidArgumentError: Found 'fast=True' in kwargs, but push_mode=WriterPushMode.Truncate is already set
```

```python
qdbnp.write_arrays(data, conn, ndex=idx, _async=True, fast=True)
# InvalidArgumentError: Found '_async=True' in kwargs, but 'fast' is already set. Please set only one, or use 'push_mode' instead
```

Verbose type checks:
```python
qdbnp.write_arrays(data, conn, index=idx, _async=63)
# InvalidArgumentError: Invalid '_async' type, expected: <class 'bool'>, got: <class 'int'>
```

```python
qdbnp.write_arrays(data, conn, "demo", index=idx, push_mode="fast")
# InvalidArgumentError: Invalid 'push_mode' type, expected: <class 'quasardb.quasardb.WriterPushMode'>, got: <class 'str'>
```

quasardb/numpy/__init__.py
```diff
@@ -705,20 +733,40 @@ def write_arrays(
...
+      * `Truncate`: Truncate (also referred to as upsert) the data in-place. Will detect time range
+        to truncate from the time range inside the dataframe.
+      * `Async`: Uses asynchronous insertion API where commits are buffered server-side and
+        acknowledged before they are written to disk. If you insert to the same table from
+        multiple processes, setting this to True may improve performance.
+      * `Fast`: Whether to use 'fast push'. If you incrementally add small batches of data to table,
+        you may see better performance if you set this to True.
+      * `Transactional`: Ensures full transactional consistency.
+
+      Defaults to `Transactional`.
```
solatis, line 748: I would make this explicit by setting push_mode=quasardb.WriterPushMode.Transactional in the function declaration

quasardb/numpy/__init__.py
```diff
@@ -747,32 +795,53 @@ def write_arrays(
...
+        if kwarg_value:
+            if push_mode and push_mode != mode:
+                raise quasardb.quasardb.InvalidArgumentError(
+                    f"Found '{kwarg}=True' in kwargs, but push_mode={push_mode} is already set"
+                )
+            if selected_push_mode:
+                raise quasardb.quasardb.InvalidArgumentError(
+                    f"Found '{kwarg}=True' in kwargs, but '{selected_kwarg}' is already set. "
+                    "Please set only one, or use 'push_mode' instead"
+                )
```
solatis, line 832: I don't understand the need for this? Wouldn't a check for `if push_mode` and updating the push_mode instead be more concise? 

I don't understand exactly what case this covers...

quasardb/numpy/__init__.py
```diff
@@ -747,32 +795,53 @@ def write_arrays(
...
+            _kwarg_deprecation_warning(kwarg, ["push_mode"], [mode], 2)
+
+    if not push_mode:
+        push_mode = selected_push_mode or quasardb.WriterPushMode.Transactional
 
     # Create batch column info from dataframe
     if writer is None:
         writer = cluster.writer()
 
+    ret = []
```
solatis, line 843: What is this doing here?

quasardb/numpy/__init__.py
rodp63: If we set the value of `push_mode`, then we won't be able to use the `_async`, `fast` and `truncate` kwargs, since we prioritize `push_mode` over the others.
However, once we remove the support for `_async`, `fast` and `truncate`, we should set this default value. I can add a `TODO` comment to not forget this.

quasardb/numpy/__init__.py
rodp63: The covered case is when we set 2 deprecated kwargs (e.g. `fast` and `_async` at the same time), it's a matter of displaying a proper error message. However, we can have a more general message to do fewer verifications.

quasardb/numpy/__init__.py
rodp63: The declaration has been moved to keep all the `push_mode` verifications at the top. No harm in this, but keeping the declarations together

quasardb/numpy/__init__.py
solatis: I'd prefer less code in this case, keep it simple.

quasardb/numpy/__init__.py
solatis: I don't understand, I don't see this variable being used anywhere?

solatis:
Could you also adjust the tests to replace any use of "async" or "fast" or "truncate" parameters with the push_mode variant?

solatis (approved):

PR #87, opened 2025-04-03
QDB-16279 - Avoid side-effects for numpy write_arrays data argument

solatis:
As reported, if the user invokes the function multiple times with the exact same arguments, the function may throw an error as it modifies the `data` dict in-place. This is not great.

This PR makes a shallow copy of the dict before `pop`'ing the item, which avoids this situation.


terngkub (approved):
LGTM

PR #90, opened 2025-04-26
QDB-16178 - Refactor statistics

alivasples:


quasardb/stats.py
```diff
@@ -147,6 +149,29 @@ def stat_type(stat_id):
...
+      Number of keys to retrieve.
+    """
+    xs = dconn.prefix_get(stats_prefix, n)
+    increase_rate = 8
+    # keep getting keys while number of results exceeds the given "n"
+    while len(xs) >= n:
+        n = n * increase_rate
+        if n >= MAX_KEYS:
+            raise Exception(f"ERROR: Cannot fetch more than {MAX_KEYS} keys.")
+        xs = dconn.prefix_get(stats_prefix, n)
```
solatis, line 170: Would be more elegant / less error-prone if you could avoid repeating the `dconn.prefix_get` statement more than once

alivasples: Sure, done :)

solatis:
/review
--pr_reviewer.require_soc2_ticket=true

PR #93, opened 2025-05-06
QDB-16178 - Refactor python statistics module

solatis:


quasardb/stats.py
```diff
@@ -1,14 +1,21 @@
 import re
+from time import sleep
```
rodp63, line 2: `sleep` is not accessed

quasardb/stats.py
```diff
@@ -144,66 +98,193 @@ def stat_type(stat_id):
...
+
+class Unit(Enum):
+    NONE = 0
+    COUNT = 1
+
+    # Size units
+    BYTES = 32
+
+    # Time/duration units
+    EPOCH = (64,)
```
rodp63, line 151: Why is this value a `tuple`?

rodp63 (approved):
2 minor comments, LGTM

PR #99, opened 2025-06-06
Fix pytest breaking changes

igorniebylski:
Since pytesty 8.4.0 tests will fail, instead of raising a warning, if they return any value other than None.

https://docs.pytest.org/en/stable/changelog.html#pytest-8-4-0-2025-06-02

Issue was identified in `tests/test_convert.py` tests.
To fix:
* don't return True for `dtype.char == "S"` edge case
* don't return _test_array_recode - assertion happens inside this function already, no need to return anything


solatis (approved):
Make sure to test it builds everywhere before merging.

igorniebylski:
other than flaky `stats_by_node` test everything is ok :)

PR #101, opened 2025-06-23
QDB-16881 - Exception when reading empty table with qdbpd.read_dataframe

igorniebylski:
Story details: https://app.shortcut.com/quasardb/story/16881


solatis (commented):
Left comment, would be good to clarify upon that.

PR #113, opened 2026-03-25
QDB-18722 - [python] - remove deprecated API code

vikonix:


examples/tutorial/pandas_tutorial.py
```diff
@@ -28,7 +28,9 @@
 
     # bulk-read-start
 
-    ranges = [(np.datetime64("2019-02-01", "ns"), np.datetime64("2019-02-02", "ns"))]
+    ranges = np.array(
+        [(np.datetime64("2019-02-01", "ns"), np.datetime64("2019-02-02", "ns"))]
+    )
```
solatis: Style changes again

examples/tutorial/pandas_tutorial.py
```diff
@@ -39,24 +41,23 @@
...
+            "close": pd.Series(data=[3.50, 3.55], index=timestamps, dtype=np.float64),
+            "volume": pd.Series(data=[10000, 7500], index=timestamps, dtype=np.int64),
+        }
+    )
 
-    # We use the write_series function to insert column by column.
-    qdbpd.write_series(opens, t, "open")
-    qdbpd.write_series(closes, t, "close")
-    qdbpd.write_series(volumes, t, "volume")
+    qdbpd.write_dataframe(df, c, t)
```
solatis: Do these tests for series really need to be ported? I'd assume that most of these are already covered in the other tests?

solatis (changes requested):
Still doing too much in a single PR, and I think you're still mistakenly trying to support the old tests / documentation / tutorials, where (better) equivalents / tests / documentation already exists for the regular bulk reader / batch writer, so you're duplicating that.

examples/tutorial/pandas_tutorial.py
solatis: I think arrays are already implicitly castable between np array and regular python lists?

solatis (changes requested):

PR #115, opened 2026-04-07
QDB-18722 - [python] - remove deprecated API code - fix bugs

vikonix:


quasardb/convert/array.hpp
```diff
@@ -186,17 +186,17 @@ struct convert_array<From, To>
...
+                return To::null_value();
             }
             else
             {
-                return static_cast<delegate_value_type>(x);
+                return static_cast<value_type>(x);
             }
         };
-        return ranges::views::transform(xform) | delegate();
+        return delegate() | ranges::views::transform(xform);
```
solatis: What is being fixed here? This appears to change 'A -> B' conversion into 'B -> A', is that intentional? Why?

quasardb/masked_array.hpp
```diff
@@ -500,11 +500,25 @@ class masked_array
...
+            for (auto cur = begin; cur != end; ++cur, ++p_ret)
+            {
+                *p_ret = (*cur == nullptr) || (*cur == Py_None);
+            }
+
+            return mask::of_array(py::cast<py::array>(ret));
+        }
+
+        using value_type = typename Dtype::value_type;
+
```
solatis: :o this does a huge array copy, right? Why can't `PyObject *` be interpreted as `py::object` -- `py::object` *is* PyObject. 

This seems like a very expensive conversion. Why is it not integrated into the `convert` facilities, and why isn't it explained better *why* we need to do this conversion here? 

In the current shape, this requires a much better approach. From what I understand you're sending a number array of objects, which we don't really support / STRONGLY DISCOURAGE, to prevent people from doing extremely performance-terrible stuff, and handle all the sanitization and conversion in `numpy/__init__.py`, such that we *don't* have to do these kind of crazy things: we support unicode and blobs both as "raw", continuous allocations. This is a conscious performance vs memory consumption tradeoff.

solatis (changes requested):
Mostly concerned about starting to support raw object conversions in our c++ code, I don't think that's a great approach. 

What exactly is the problem with binary data that needs to be fixed? 

Doing this special-case conversion directly inside the masked array is the wrong place, if you implement this in convert/value.hpp (but.. i think you already had added a py::object based conversion there?), all array, masked array, etc conversions will automatically make use of that.

quasardb/masked_array.hpp
vikonix: This special-case is needed because NumPy object arrays store raw PyObject * values, not py::object wrappers. The generic masked_null() path reads elements as Dtype::value_type, which is valid for regular fixed-width dtypes but not for traits::pyobject_dtype.

For dtype('O'), we therefore read the buffer as PyObject * const * and mark entries as null when the slot is nullptr or Py_None.

This does not introduce an extra object-array copy. masked_null() already allocates a boolean mask and scans the array; this only changes how object-array elements are read during that scan. The fix belongs here because the bug was in mask reconstruction, not in the conversion step that produced the object array.

PR #117, opened 2026-04-10
QDB-18722 - add Read Arrays

vikonix:


vikonix:
Added the Numpy ReadArrays

Added the writer access from table class
This needs for implementing a compatibility call WriteArray before we fully remove this API

quasardb/numpy/__init__.py
```diff
@@ -705,6 +705,169 @@ def write_array(
...
+        )
+
+    if len(idx_batches) == 1:
+        return idx_batches[0], {
+            cname: value_batches[cname][0] for cname in column_names
+        }
+
+    return np.concatenate(idx_batches), {
+        cname: _concat_masked(batches) for (cname, batches) in value_batches.items()
+    }
```
solatis, line 761: This concatenates all batches, which defeats the whole purpose of the streaming aspect, not?

solatis: Should yield individual batches instead

quasardb/numpy/__init__.py
```diff
@@ -705,6 +705,169 @@ def write_array(
...
+
+
+def read_arrays(
+    cluster: Optional[quasardb.Cluster] = None,
+    table: Optional[Union[str, Table]] = None,
+    columns: Optional[Sequence[str]] = None,
+    ranges: Any = None,
+) -> Tuple[NDArrayTime, Dict[str, MaskedArrayAny]]:
```
solatis: Should return a generator

quasardb/table.cpp
```diff
@@ -263,4 +264,10 @@ qdb::reader_ptr table::reader(                     //
     return std::make_unique<qdb::reader>(_handle, table_names, column_names, batch_size, ranges);
 };
 
+qdb::writer_ptr table::writer() const
+{
+    _handle->check_open();
+    return std::make_unique<qdb::writer>(_handle);
+}
+
```
solatis: Why is this necessary? A writer operates on multiple tables

solatis (changes requested):
This is not the right implementation. You should have taken the bulk reader from the pandas API (with streaming support that yields individual batches, because that's the whole reason the bulk reader exists -- streaming over a table rather than ingesting everything in memory).

Then the pandas adapter wraps the numpy adapter, like it does with all other APIs, and we don't need to have two different consumers of the bulk reader. Then it mirrors the approach of the writer and our old column-oriented API. I'm not sure why the bulk reader was implemented in the pandas adapter, that must have been an oversight.

solatis (approved):

PR #118, opened 2026-04-13
QDB-18722 - [python] - remove deprecated API code

vikonix:


tests/test_numpy.py
```diff
@@ -28,48 +28,79 @@ def _unicode_to_object_array(xs):
...
+        payload,
+        conn,
+        table,
+        index=index,
+        dtype=dtype,
+        infer_types=infer_types,
+        **kwargs,
+    )
+
+
```
solatis, line 58: Are these necessary, are you adding backawards compatibility for single column tests that were originally testing the old API?

solatis (commented):
Seems like you're adding backwards compatibility to tests that can be removed?

tests/test_numpy.py
vikonix: no, this wrappers need for tests that operate just with one column
we have a lot of them

vikonix: tests check different data conversions

vikonix: for example, `test_arrays_deduplicate`

vikonix: test_arrays_read_write_inferrable_dtypes

solatis (approved):

PR #123, opened 2026-06-22
QDB-19269 - [python] Fix API and tests for bulk reader column order

vikonix:
QDB-19269 - [python] Fix API and tests for bulk reader column order

### DIFF
--- a/quasardb/pandas/__init__.py
+++ b/quasardb/pandas/__init__.py
@@ -203,7 +203,7 @@ def stream_dataframe(
     # For backwards compatibility, we drop the `$table` column returned: this is not strictly
     # necessary, but it also is somewhat reasonable to drop it when we're reading from a single
     # table, which is the case here.
-    # When a column list is explicitly provided, `$table` may not be in that list.
+    # When a column list is explicitly provided, `$table` may not be in that list.    
     clean_df_fn = lambda df: (
         df.drop(columns=["$table"]) if "$table" in df.columns else df
     )
--- a/quasardb/pandas/__init__.py
+++ b/quasardb/pandas/__init__.py
@@ -203,10 +203,7 @@ def stream_dataframe(
     # For backwards compatibility, we drop the `$table` column returned: this is not strictly
     # necessary, but it also is somewhat reasonable to drop it when we're reading from a single
     # table, which is the case here.
-    # When a column list is explicitly provided, `$table` may not be in that list.
-    clean_df_fn = lambda df: (
-        df.drop(columns=["$table"]) if "$table" in df.columns else df
-    )
+    clean_df_fn = lambda df: df.drop(columns=["$table"]) if not column_names else df
 
     return (
         clean_df_fn(df)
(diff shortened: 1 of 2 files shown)


quasardb/pandas/__init__.py
```diff
@@ -203,7 +203,8 @@ def stream_dataframe(
     # For backwards compatibility, we drop the `$table` column returned: this is not strictly
     # necessary, but it also is somewhat reasonable to drop it when we're reading from a single
     # table, which is the case here.
-    clean_df_fn = lambda df: df.drop(columns=["$table"])
+    # When a column list is explicitly provided, the reader may not return `$table`.
+    clean_df_fn = lambda df: df.drop(columns=["$table"], errors="ignore")
```
solatis: Wouldn't it be better to not invoke drop if $table doesn't exist? It seems a bit unintuitive to fix it this way.

quasardb/pandas/__init__.py
```diff
@@ -203,7 +203,8 @@ def stream_dataframe(
     # For backwards compatibility, we drop the `$table` column returned: this is not strictly
     # necessary, but it also is somewhat reasonable to drop it when we're reading from a single
     # table, which is the case here.
-    clean_df_fn = lambda df: df.drop(columns=["$table"])
+    # When a column list is explicitly provided, the reader may not return `$table`.
```
solatis: Why "may" not ? That means sometimes it does ?

vikonix: now if the user explicitly specifies columns, only they will be returned
if column list is empty then we return $table column as previously

quasardb/pandas/__init__.py
vikonix: fixed

quasardb/pandas/__init__.py
vikonix: fixed

quasardb/pandas/__init__.py
```diff
@@ -203,7 +203,10 @@ def stream_dataframe(
     # For backwards compatibility, we drop the `$table` column returned: this is not strictly
     # necessary, but it also is somewhat reasonable to drop it when we're reading from a single
     # table, which is the case here.
-    clean_df_fn = lambda df: df.drop(columns=["$table"])
+    # When a column list is explicitly provided, `$table` may not be in that list.
+    clean_df_fn = lambda df: (
+        df.drop(columns=["$table"]) if "$table" in df.columns else df
+    )
```
solatis: I'm still not a super big fan of the comment and logic here. 

What is the new behavior of the bulk reader? $table is only returned if it is part of the provided `column_names`?

Because this logic here is implemented for backwards compatibility. It strikes me that this whole clean function is now not necessary anymore? Because the new behavior is as it used to be? 

Right now we even drop $table if it was part of the requested columns, which I don't think is the intention.

vikonix: $table returned:
1. column list not provided then we return full column list including $table (as previously)
2. column list provided we return only specified columns

solatis (approved):

PR #127, opened 2026-07-22
QDB-15924 - Expose lazy table creation mode in batch writer to Python API - 2

vikonix:


examples/tutorial/pandas_tutorial.py
```diff
@@ -19,10 +19,21 @@
...
+            quasardb.ColumnInfo(quasardb.ColumnType.Double, "close"),
+            quasardb.ColumnInfo(quasardb.ColumnType.Int64, "volume"),
+        ],
+    )
+    qdbpd.write_dataframe(
+        df,
+        c,
+        "stocks",
+        create_schemas={"stocks": stocks_schema},
+    )
```
solatis: I think this is not a great example here, as this particular example already created the table. This should be a separate example.

quasardb/detail/writer.cpp
```diff
@@ -207,6 +207,27 @@ std::vector<qdb_exp_batch_push_column_t> const & staged_table::prepare_columns()
     return _columns_data;
 }
 
+std::vector<qdb_exp_batch_push_column_schema_t> const & staged_table::prepare_column_schema()
+{
+    _column_schema.clear();
+    _column_schema.reserve(_column_infos.size());
```
solatis: Why do you call this `column schema` but the outer API `table schema` ? aren't they the same?

quasardb/detail/writer.cpp
```diff
@@ -217,10 +238,23 @@ void staged_table::prepare_table_data(qdb_exp_batch_push_table_data_t & table_da
...
+{
+    table_schema.shard_size = static_cast<qdb_duration_t>(_shard_size.count());
+    table_schema.ttl        = (_ttl == std::chrono::milliseconds::zero())
+                                  ? qdb_ttl_disabled
+                                  : static_cast<qdb_duration_t>(_ttl.count());
+
+    const auto & columns      = prepare_column_schema();
+    table_schema.columns      = columns.data();
+    table_schema.column_count = columns.size();
+}
```
solatis: Why is `_column_schema` stored as part of the staged table object, but table schema not? 

(I know the answer, but this should be documented / commented)

quasardb/detail/writer.hpp
```diff
@@ -160,6 +160,9 @@ class staged_table
     staged_table(qdb::table const & table)
         : _logger("quasardb.writer")
         , _table_name(table.get_name())
+        , _creation_allowed(table.has_local_schema())
+        , _shard_size(table.get_shard_size())
+        , _ttl(table.get_ttl())
```
solatis: Aren't `shard size` and `ttl` part of the table schema? Shouldn't we potentially refactor the way this is represented, rather than have a lot of different class parameters on staged table? I.e. keep the schema in a separate struct.

Also: why do we store shard size and ttl as part of the constructor, but not the columns? Are you sure this is the place where we should store it?

quasardb/detail/writer.hpp
```diff
@@ -227,6 +235,7 @@ class staged_table
         _table_name.clear();
         _column_infos.clear();
         _columns_data.clear();
+        _column_schema.clear();
```
solatis: I'm not a fan of "column schema" as name, it seems to conflict with "column infos", it's unclear what it is.

quasardb/detail/writer.hpp
```diff
@@ -256,6 +269,12 @@ class staged_table
     std::vector<any_column> _columns;
 
     std::vector<qdb_exp_batch_push_column_t> _columns_data;
+
+    // Only for lazy table creation, we need to provide the column schema to the server.
+    const bool _creation_allowed;
+    std::vector<qdb_exp_batch_push_column_schema_t> _column_schema;
+    std::chrono::milliseconds _shard_size;
+    std::chrono::milliseconds _ttl;
 };
```
solatis, line 269: I think we need to reuse the "table schema" concept here, so that we can consistently represent the table schema as the same class / object.

quasardb/pandas/__init__.py
```diff
@@ -336,39 +310,74 @@ def write_dataframes(
...
+                        type(schema_alias)
+                    )
+                )
+            if not isinstance(configured_schema, quasardb.TableSchema):
+                raise quasardb.InvalidArgumentError(
+                    (
+                        "Invalid 'create_schemas' value type, expected: "
+                        "TableSchema, got: {}"
+                    ).format(type(configured_schema))
+                )
```
solatis: I think this should all be pass-through to the numpy API, qdbnp. We want that API to have the same capabilities, i.e. I can also make use of the auto-create functionality if I use the numpy api. The pandas api just passes things through.

quasardb/pandas/__init__.py
```diff
@@ -336,39 +310,74 @@ def write_dataframes(
...
-
-        cinfos = [(x.name, x.type) for x in table.list_columns()]
+        table_alias = table if isinstance(table, str) else table.get_name()
+        schema = create_schemas.get(table_alias) if create_schemas is not None else None
+        if schema is not None:
+            cinfos = [(column.name, column.type) for column in schema.columns]
+        else:
+            if isinstance(table, str):
+                table = table_cache.lookup(table_alias, cluster)
+            cinfos = [(x.name, x.type) for x in table.list_columns()]
```
solatis: What happens here if schema does not match the actual schema? Did you add tests for this?

quasardb/table.hpp
```diff
@@ -49,12 +62,34 @@ class table : public entry
...
+     * The schema, shard size, and TTL are assumed to match an existing table with the same alias.
+     */
+    table(handle_ptr h,
+        std::string a,
+        std::vector<detail::column_info> columns,
+        std::chrono::milliseconds shard_size,
+        std::chrono::milliseconds ttl)
+        : entry{h, a}
+        , _has_indexed_columns(false)
+        , _has_local_schema(true)
```
solatis: I think we should represent the table schema as a std::optional<table::schema> or whatever, then we don't need the `has local schema` anymore.

quasardb/table.hpp
```diff
@@ -49,12 +62,34 @@ class table : public entry
...
+        , _columns{normalize_local_schema(std::move(columns))}
+        , _ttl{ttl}
+        , _shard_size{shard_size}
```
solatis: Should be stored as a `table::schema`

quasardb/table_schema.py
```diff
@@ -0,0 +1,39 @@
...
+from quasardb.quasardb import ColumnInfo
+
+
+@dataclass
+class TableSchema:
+    """Schema used to create a missing time-series table during a writer push."""
+
+    columns: List[ColumnInfo]
+    shard_size: timedelta = timedelta(days=1)
+    ttl: timedelta = timedelta(0)
```
solatis: What's the purpose of this? Isn't the standard in the API that we define these classes in C++, if they're used from C++? 

If this is *not* used from C++, perhaps we should?

quasardb/writer.hpp
```diff
@@ -197,25 +198,55 @@ class writer
...
-            detail::staged_table & staged_table = pos->second;
-            auto & batch_table                  = batch.at(cur++);
+            std::string const & table_name                    = pos->first;
+            detail::staged_table & staged_table               = pos->second;
+            auto & batch_table                                = batch.at(cur);
+            qdb_exp_batch_creation_mode_t table_creation_mode = qdb_exp_batch_dont_create;
+
+            if (creation_mode == qdb_exp_batch_create_tables)
+            {
+                if (staged_table.creation_allowed() == false)
```
solatis: I would add [[unlikely]] here

quasardb/writer.hpp
```diff
@@ -197,25 +198,55 @@ class writer
...
         {
-            std::string const & table_name      = pos->first;
-            detail::staged_table & staged_table = pos->second;
-            auto & batch_table                  = batch.at(cur++);
+            std::string const & table_name                    = pos->first;
+            detail::staged_table & staged_table               = pos->second;
+            auto & batch_table                                = batch.at(cur);
+            qdb_exp_batch_creation_mode_t table_creation_mode = qdb_exp_batch_dont_create;
+
+            if (creation_mode == qdb_exp_batch_create_tables)
```
solatis: can't "creation mode" and "creation allowed" be folded into a single argument? 

i.e. if a schema is provided, creation is enabled, if not, it's not enabled. 

will simplify things

PR #128, opened 2026-07-22
QDB-15924 - fix bug

vikonix:


solatis (approved):
Good catch, please change the PR title to be a bit more descriptive though :)

PR #129, opened 2026-07-24
QDB-19316 - Fix Python API after removing old ts_batch interfaces

K-ballo:


examples/tutorial/python.py
```diff
@@ -79,41 +80,21 @@ def secure_connect():
...
-    inserter.set_double(1, 3.55)
-    inserter.set_int64(2, 7500)
+    index = np.array(
+        [np.datetime64("2019-02-01", "ns"), np.datetime64("2019-02-02", "ns")]
+    )
 
-    # Now that we're done, we push the buffer as one single operation.
-    inserter.push()
+    # Now that we're done, we push the arrays as one single operation.
+    qdbnp.write_arrays(data, c, t, index=index, infer_types=False)
```
solatis, line 97: Just noting: technically speaking, this should use the "raw" batch inserter API, but that's a very "unfriendly" API not actually designed for direct invocation. What you're describing now is the numpy API, I think this is fine.

solatis (approved):
lgtm!

